PR #27, opened 2020-03-27
Automate publishing of cli

AndrewSpeed:
# What changed?

This PR adds the ability into the CircleCI config for us to publish the CLI, with a guard to prevent us from publishing on each merge into master.

We can then use the `bin/publish` script to trigger the pipeline, overriding the `should_publish` parameter, pubishing to pypi and cloudsmith.

## To Do
 - [ ] Add appropriate environment variables to the pipeline settings

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,12 @@
 #
 version: 2.1
 
+parameters:
+  should_publish:
+    description: Whether the CLI should be published to repositories
+    type: boolean
+    default: false
+
 jobs:
   publish:
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -4,6 +4,37 @@
...
+        description: Twine password
+        type: string
+    environment:
+      TWINE_REPOSITORY_URL: <<parameters.registry_url>>
+      TWINE_USERNAME: <<parameters.username>>
+      TWINE_PASSWORD: <<parameters.password>>
+    steps:
+      - checkout
+      - run: pip install --user --upgrade setuptools twine wheel
+      - run: setup.py sdist bdist_wheel
```
lskillen: Should be `--universal` since we're building for py2 and py3. :-)

lskillen: Nevermind. We've got it in the `setup.cfg`. Picks it up, right?

bin/publish
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
lskillen: If we did this based on a tag being pushed, that'd get rid of the need to call `bin/publish` - Right?

MrTam (approved):

PR #35, opened 2020-07-24
Add asciinema video

DanMcKinney:
Added a short asciinema video to demo some CLI commands, also added the demo.cast file itself.


README.md
```diff
@@ -11,9 +11,10 @@
 
 The [Cloudsmith](https://cloudsmith.io) Command Line Interface (CLI) is a Py2/Py3 text-based interface to the [API](https://api.cloudsmith.io). This allows users, machines and other services to access and integrate smoothly with Cloudsmith without requiring explicit plugins or tools. [Be awesome. Automate Everything](https://cloudsmith.com/company/the-tao-of-cloudsmith/).
 
-The following (very out of date) GIF demonstrates a small slice of the CLI - [View the full video on YouTube](https://youtu.be/8nlstYU9J5M):
+The following asciienema video demonstrates some of the CLI commands:
+[![asciicast](https://asciinema.org/a/DkNXQWQGBjWkfGPAkDAPNz7xe.svg)](https://asciinema.org/a/DkNXQWQGBjWkfGPAkDAPNz7xe)
```
apoclyps, line 15: looks like the video is uploaded to a personal account; should this be on a cloudsmith account instead?

DanMcKinney: This is a good point - the account is under dmckinney@cloudsmith.io

DanMcKinney: Account updated to "Cloudsmith" - still pulls my avatar from elsewhere though

README.md
```diff
@@ -11,9 +11,10 @@
 
 The [Cloudsmith](https://cloudsmith.io) Command Line Interface (CLI) is a Py2/Py3 text-based interface to the [API](https://api.cloudsmith.io). This allows users, machines and other services to access and integrate smoothly with Cloudsmith without requiring explicit plugins or tools. [Be awesome. Automate Everything](https://cloudsmith.com/company/the-tao-of-cloudsmith/).
 
-The following (very out of date) GIF demonstrates a small slice of the CLI - [View the full video on YouTube](https://youtu.be/8nlstYU9J5M):
+The following asciienema video demonstrates some of the CLI commands:
```
MrTam, line 14:
```suggestion
The following asciinema video demonstrates some of the CLI commands:
```

README.md
apoclyps: will others have access to the shared account to upload also?

demo.cast
```diff
@@ -0,0 +1,565 @@
+{"version": 2, "title": "Cloudsmith CLI", "width": 194, "height": 74, "timestamp": 1595593770, "env": {"SHELL": "/usr/local/bin/bash", "TERM": "xterm-256color"}}
```
apoclyps, line 1: we can probably skip uploading this file; I'm not sure there is much we can do to reuse them other than to upload it again.

README.md
DanMcKinney: lol great spot. too many (or too little) "i"s

MrTam: I can see why! 😄

lskillen:
Would recommend embedding as an image instead of as a link alone:
https://asciinema.org/docs/embedding

README.md
DanMcKinney: They will need to, so I'll need to change the email address on the account to engineering@ or similar - no passwords on asciinema, all logins done by one time links sent via email. Suggestions for the email account to use?

MrTam:
> Would recommend embedding as an image instead of as a link alone:
> https://asciinema.org/docs/embedding

I believe it already is, using the markdown example.

DanMcKinney:
Sure thing - just used the markdown from the docs as it seemed most like the rest of the README. I've changed it to an embedded image now.

README.md
apoclyps: `engineering@cloudsmith.io`probably makes the most sense to avoid additional setup. we can always change it later if it gets noisy

DanMcKinney: Shouldn't get too bad, I'll change that now.

DanMcKinney:
Yeah actually I think Tom is right, I think it's an embedded svg from the markdown example in the docs.

README.md
DanMcKinney: Account name updated to "Cloudsmith" account email set to engineering@cloudsmith.io - need a one time link sent to that address to log in 👍 .

PR #46, opened 2020-11-18
Organization usage metrics

apoclyps:
### What's Changed

Updates command to display organization usage metrics for a namespace e.g.

```bash
> cloudsmith metrics tokens ORG


Getting usage metrics ... OK

Token Usage Totals
---------------------------------------------------------------
Total Tokens | Active Tokens | Inactive Tokens | Bandwidth Used
58           | 56            | 2               | 15.8 TB

Bandwidth Usage Statistics
---------------------------------------------------------------
Lowest usage | Average usage | Highest usage
0 B          | 2.3 GB        | 445.3 GB
```

**Existing usage**

```
> cloudsmith metrics tokens ORG/REPO


Getting usage metrics ... OK

Token Usage Totals
---------------------------------------------------------------
Total Tokens | Active Tokens | Inactive Tokens | Bandwidth Used
13           | 11            | 2               | 22.8 GB

Bandwidth Usage Statistics
---------------------------------------------------------------
Lowest usage | Average usage | Highest usage
562 B        | 20.6 MB       | 1.3 GB
```

**Missing required args:**

```bash
> cloudsmith metrics tokens

Usage: cloudsmith metrics tokens [OPTIONS] OWNER or OWNER/REPO

Error: Invalid value for "OWNER or OWNER/REPO": Individual values cannot be blank
```


cloudsmith_cli/core/api/metrics.py
```diff
@@ -14,6 +14,29 @@ def get_metrics_api():
     return get_api_client(cloudsmith_api.MetricsApi)
 
 
+def organization_entitlement_usage_metrics(owner=None, **kwargs):
+    """List token usage metrics."""
+    client = get_metrics_api()
+
+    api_kwargs = {}
+    api_kwargs.update(
```
MrTam, line 22: NAB: Could just use comprehension below as initialiser rather than declare a new dict and update?

cloudsmith_cli/core/api/metrics.py
```diff
@@ -23,10 +46,13 @@ def entitlement_usage_metrics(owner=None, repo=None, **kwargs):
...
     )
 
+    headers = []
+    res = None
+
     if owner and repo:
         if hasattr(client, "metrics_entitlements_usage_list_with_http_info"):
             with catch_raise_api_exception():
-                res, _, headers = client.metrics_entitlements_usage_list_with_http_info(
+                res, _, headers = client.metrics_entitlements_usage_list0_with_http_info(
```
MrTam, line 55: Is this a typo, with the 0?

MrTam (approved):

cloudsmith_cli/cli/validators.py
```diff
@@ -100,6 +100,12 @@ def validate_optional_owner_repo(ctx, param, value):
     )
 
 
+def validate_required_owner_optional_repo(ctx, param, value):
+    """Ensure that owner/repo is formatted correctly, where owner is required and repo is optional."""
+    form = "OWNER/[/REPO]"
```
lskillen, line 105: Wee typo here (double-slash).

cloudsmith_cli/core/api/metrics.py
apoclyps: nope, auto-generated bindings add a prefix of `0` which is super annoying - we reuse the base serializer for org metrics and repo metrics which is why we have this change here for the existing endpoint.

cloudsmith_cli/core/api/metrics.py
apoclyps: I agree it would be cleaner however this is the pattern that's been used across multiple commands. If we were to change it, it should be updated everywhere for consistency but there is probably a bigger refactoring pass that is needed for the CLI that this could be part of.

PR #57, opened 2021-10-09
Skip click 8.0.2

dylwylie:
click 8.0.2 has a bug where it mangles the passed in filename and causes
validation to fail with "file not found"


dylwylie:
relates to https://github.com/pallets/click/issues/2088

PR #60, opened 2021-10-11
skip click-8.0.2 but allow patched versions

sanderr:
A couple of days ago the `click` dependency was constrained to `<8.0.2` because of a bug in that version (#57). The bug has since been fixed and released as `click-8.0.3` (pallets/click#2088, pallets/click#2094, https://click.palletsprojects.com/en/8.0.x/changes/#version-8-0-3). I believe the constraint could now be loosened to only disallow 8.0.2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
     zip_safe=False,
     platforms=["any"],
     install_requires=[
-        "click>=7.0,<8.0.2",
+        "click>=7.0,!=8.0.2",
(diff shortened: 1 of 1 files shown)


sanderr:
Thanks for the pointer. I've set it up but I'm not sure how to continue now. Do I understand correctly that all commits in this pull request need to be signed and that I can't go back and sign commits I've already pushed? Should I sign the commit locally and force push it then?

PR #67, opened 2021-12-08
Allow cloudsmith-api <2

paddycarey:
Relax API restrictions to allow installation of newer versions of our API clients.


lskillen (approved):
:+1:

PR #68, opened 2021-12-10
Fixed output values to align with correct headers

DanMcKinney:
While the values for quota usage were correct, they were not in the right columns, so just a simple correction!

### DIFF
--- a/cloudsmith_cli/cli/commands/quota/quota.py
+++ b/cloudsmith_cli/cli/commands/quota/quota.py
@@ -45,9 +45,9 @@ def display_quota(opts, data):
     rows = []
     rows.append(
         [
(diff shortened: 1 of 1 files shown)


dbmurf93:
[failed pytest](https://app.circleci.com/pipelines/github/cloudsmith-io/cloudsmith-cli/309/workflows/23774f8d-4fbd-429f-9060-253878b7cd39/jobs/1406) but seems like it was an issue with circle that caused it

dbmurf93:
tested and appears to be working, though
## Screenshots
_Before_
image

_After_
image

dbmurf93:
Closed this one in lieu of clone

PR #74, opened 2022-03-03
fix: support new swagger-codegen format for cloudsmith-api

paddycarey:
This PR adjusts the CLI to support `swagger-codegen`'s newer `Configuration()` style (without the global singleton).

Upstream change was made here https://github.com/swagger-api/swagger-codegen/pull/9130

This change is required due to an upgrade in the `swagger-codegen` version we use to generate our API client libs over in https://github.com/cloudsmith-io/cloudsmith-api/pull/27

### DIFF
--- a/cloudsmith_cli/core/api/init.py
+++ b/cloudsmith_cli/core/api/init.py
@@ -48,7 +48,11 @@ def initialise_api(
             values = decoded.decode("utf-8")
             config.username, config.password = values.split(":")
 
-    set_api_key(config, key)
+    if key:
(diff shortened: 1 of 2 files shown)


paddycarey:
> Should this also update https://github.com/cloudsmith-io/cloudsmith-cli/blob/master/setup.py#L55 to exclude incompatible API versions?

The CLI remains compatible with both old and new versions of the API bindings after this change, so no need to update here.

DavidS-cloud (approved):

PR #79, opened 2022-05-11
Fix config homedir

dmeecs:
Per the readme, the `~/.cloudsmith` should be searched for .ini files as well.


cloudsmith_cli/cli/config.py
```diff
@@ -45,7 +45,14 @@ def get_error_hint(self, ctx):
 
 def get_default_config_path():
     """Get the default path to cloudsmith config files."""
-    return click.get_app_dir("cloudsmith")
+    return click.get_app_dir("cloudsmith")  # only returns a single path
+
+
+_CFG_SEARCH_PATHS = (
+    ".",
+    os.path.expanduser("~/.cloudsmith"),
```
paddycarey, line 53: what does this look like on windows?

dmeecs: It'll be pretty similar, per my reading of https://docs.python.org/3/library/os.path.html#os.path.expanduser 

So it would be something like `C:\users\dmee\.cloudsmith`. Which the readme doesn't include in the window specific paths. I would be tempted to update the readme to include `C:\users\dmee\.cloudsmith\` as a valid config location for windows as I've certainly dropped lots of config files there before.

dmeecs: ah wait what the readme is really getting at, but doesn't say it, is that the current behaviour was defined by this: https://click.palletsprojects.com/en/5.x/api/#click.get_app_dir   and so the 'example' folders are really exclusive, it'll only be one or the other.

paddycarey: yeah that's fair, i read it the same way

i think in the windows case we probably _don't_ want the `.` dir though, i don't believe that's very idiomatic

lskillen (approved):
Looks good to me. From:
https://www.geeksforgeeks.org/python-os-path-expanduser-method/

> On Unix platforms, an initial ~ is replaced by the value of HOME environment variable, if it is set. Otherwise, os.path.expanduser() method search for user’s home directory in password directory using an in-built module pwd. Path containing an initial ~user component is looked up directly in the password directory.

> On Windows platform, an initial ~ is replaced by the value of HOME and USERPROFILE environment variable, if it is set. Otherwise, HOMEPATH and HOMEDRIVE environment variable will be used. While Path containing an initial ~user component is handled by replacing the last directory component with ~user from the path derived above.

So it should work OK for both Linux and Windows.

cloudsmith_cli/cli/config.py
dmeecs: As in the '~/.cloudsmith' or '~/.config/cloudsmith' is not idiomatic on windows (rather than '.' as in the current working dir)? I would agree it's not idiomatic on Windows and I guess that's why the library call uses AppData as it's the idiomatic thing on windows.

paddycarey (approved):

PR #88, opened 2022-12-06
Support cloudsmith-api v2.0.0

jackgibson1:
- Python binding `v2.0.0` contains breaking changes - https://github.com/cloudsmith-io/cloudsmith-api/pull/34.
- This is due to OpenAPI specification now including serialiser fields as there actual types, rather than an object/dict. 
- Updated to access by class attributes rather than dictionary keys. 
- Updated certain areas to use the new models.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.33.0
+current_version = 0.34.0
```
jackgibson1, line 2: Updated to match the current `cli` version (tag `v0.34.0` already exists).

cloudsmith_cli/cli/commands/entitlements.py
```diff
@@ -132,6 +132,17 @@ def print_entitlements(opts, data, page_info=None, show_list_info=True):
 
     rows = []
     for entitlement in sorted(data, key=itemgetter("name")):
+        ent_updated_at = (
+            entitlement["updated_at"].strftime("%Y/%m/%d, %H:%M:%S")
+            if entitlement["updated_at"]
+            else None
+        )
+        ent_created_at = (
+            entitlement["created_at"].strftime("%Y/%m/%d, %H:%M:%S")
```
paddycarey, line 141: Is this the same format as before? Kinda feels like we should just do iso8601, but not worth breaking an interface for probably.

paddycarey (approved):
LGTM, just one question but NAB

cloudsmith_cli/cli/commands/entitlements.py
jackgibson1: Yeah it's different - shall I keep it to iso?

PR #98, opened 2023-03-29
Bump version: 0.36.1 → 0.37.0

dmeecs:


AndrewSpeed (approved):
NAB, but we should also update the changelog

PR #103, opened 2023-05-01
fix: Update generation of format parameter descriptions.

jackgibson1:
- Recent changes have contributed to a minor difference in `docstring` formatting.
- This change takes into account the new difference. 

For example: 
```
>>> attr = getattr(MavenPackageUploadRequest, "artifact_id")
>>> attr.__doc__
'Gets the artifact_id of this MavenPackageUploadRequest.\n\n        The ID of the artifact.\n\n        :return: The artifact_id of this MavenPackageUploadRequest.\n        :rtype: str\n        '
>>> docs = attr.__doc__.strip().split("\n")
>>> docs[1]
''
```

### DIFF
--- a/cloudsmith_cli/core/api/packages.py
+++ b/cloudsmith_cli/core/api/packages.py
@@ -254,7 +254,9 @@ def get_package_formats():
 
         for k, v in six.iteritems(cls.swagger_types):
             attr = getattr(cls, k)
-            docs = [doc.strip() for doc in attr.__doc__.strip().split("\n") if doc.strip()]
+            docs = [
+                doc.strip() for doc in attr.__doc__.strip().split("\n") if doc.strip()
+            ]
             doc = (docs[1] if docs[1] else docs[0]).strip()
 
             try:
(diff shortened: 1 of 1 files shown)


Leeoc (approved):

cloudsmith_cli/core/api/packages.py
```diff
@@ -255,7 +255,7 @@ def get_parameters(cls):
         for k, v in six.iteritems(cls.swagger_types):
             attr = getattr(cls, k)
             docs = attr.__doc__.strip().split("\n")
-            doc = (docs[1] if docs[1] else docs[0]).strip()
+            doc = (docs[2].lstrip() if docs[2] else docs[0]).strip()
```
chrisimcevoy, line 258: NAB: Would it be better to use a list comprehension here (or a generator expression?) which ignores blank strings?

chrisimcevoy: Something like:

```python
docs = [s.strip() for s in attr.__doc__.strip().split("\n") if s.strip()]
doc = ...
```

jackgibson1: Good idea - done 👍

chrisimcevoy (approved):

PR #105, opened 2023-05-03
Add tests for login command

chrisimcevoy:
Adds tests for `cloudsmith login`.

The technique I adopted here is to allow environment variables to drive which tests are run. For instance, any tests that depend on the presence of an api key will be skipped unless the `PYTEST_CLOUDSMITH_API_KEY` environment variable is set.

I decided to use a `PYTEST_` prefix for test-related environment variables so as not to conflict with pre-existing environment variables which are already set for actual cli usage. This way we can ensure that tests only ever target the intended environment, for example.


Leeoc (approved):
LGTM 👍

dylwylie (approved):

PR #113, opened 2023-05-10
feat(eng-2261): License policy CLI

Leeoc:
Adds functionality to manage license policies via CLI

image

### DIFF
--- a/cloudsmith_cli/cli/commands/policy/license.py
+++ b/cloudsmith_cli/cli/commands/policy/license.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""CLI/Commands - create, retrieve, update or delete licence policies."""
+"""CLI/Commands - create, retrieve, update or delete license policies."""
 from __future__ import absolute_import, print_function, unicode_literals
 
 import json
@@ -14,13 +14,13 @@ from ...utils import fmt_bool, fmt_datetime, maybe_spinner
 from .command import policy
 
 
-def print_licence_policies(policies):
-    """Print licence policies as a table or output in another format."""
+def print_license_policies(policies):
+    """Print license policies as a table or output in another format."""
 
     headers = [
         "Name",
         "Description",
-        "Allow Unknown Licences",
+        "Allow Unknown Licenses",
         "Quarantine On Violation",
         "SPDX Identifiers",
         "Created",
@@ -47,11 +47,11 @@ def print_licence_policies(policies):
     click.echo()
 
     num_results = len(rows)
-    list_suffix = "licence polic%s" % ("y" if num_results == 1 else "ies")
+    list_suffix = "license polic%s" % ("y" if num_results == 1 else "ies")
     utils.pretty_print_list_info(num_results=num_results, suffix=list_suffix)
 
 
-@policy.group(cls=command.AliasGroup, name="licence", aliases=[])
+@policy.group(cls=command.AliasGroup, name="license", aliases=[])
 @decorators.common_cli_config_options
 @decorators.common_cli_output_options
 @decorators.common_api_auth_options
@@ -59,7 +59,7 @@ def print_licence_policies(policies):
 @click.pass_context
 def licence(*args, **kwargs):
     """
-    Manage licence policies for an organization.
+    Manage license policies for an organization.
 
     See the help for subcommands for more information on each.
     """
@@ -77,7 +77,7 @@ def licence(*args, **kwargs):
 @click.pass_context
 def ls(ctx, opts, owner, page, page_size):
     """
-    List licence policies.
+    List license policies.
 
     This requires appropriate permissions for the owner (a member of the
     organisation and a valid API key).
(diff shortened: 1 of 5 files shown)


chrisimcevoy:
Looks good at first glance. I think I confused you earlier with the whole license/licence discussion. We should adhere to "International English" in anything user-facing like commands and output. Maybe it would be better to switch back to using "license" everywhere to avoid confusion, and if you have any cases where a variable named "license" shadows the builtin, then you could just append an underscore to the variable name i.e. license_

Leeoc:
> Looks good at first glance. I think I confused you earlier with the whole license/licence discussion. We should adhere to "International English" in anything user-facing like commands and output. Maybe it would be better to switch back to using "license" everywhere to avoid confusion, and if you have any cases where a variable named "license" shadows the builtin, then you could just append an underscore to the variable name i.e. license_

The only place `license` was an issue was for creating the base AliasGroup, just realised I can set the group to `license` and have the function def as `licence` to avoid the builtin... :noob:

cloudsmith_cli/cli/commands/policy/license.py
```diff
@@ -0,0 +1,324 @@
...
+    if spdx_identifiers is None:
+        raise click.BadParameter(
+            "SPDX Identifiers is a required field for creating a license policy.",
+            param="spdx_identifiers",
+        )
+
+    if len(spdx_identifiers) > len(set(spdx_identifiers)):
+        raise click.BadParameter(
+            "SPDX Identifiers must be unique.", param="spdx_identifiers"
+        )
```
chrisimcevoy, line 169: This isn't the case anymore iirc. The server will silently de-dupe them.

cloudsmith_cli/cli/commands/policy/license.py
```diff
@@ -0,0 +1,324 @@
...
+        raise click.BadParameter(
+            "Name is a required field for creating a license policy.", param="name"
+        )
+
+    spdx_identifiers = policy_config.get("spdx_identifiers", None)
+    if spdx_identifiers is None:
+        raise click.BadParameter(
+            "SPDX Identifiers is a required field for creating a license policy.",
+            param="spdx_identifiers",
+        )
```
chrisimcevoy, line 164: NAB: Would it be better UX to silently assign an empty list? Genuine question - I could go either way on that myself.

cloudsmith_cli/cli/commands/policy/license.py
```diff
@@ -0,0 +1,324 @@
...
+            "owner": click.style(owner, bold=True),
+        },
+        nl=False,
+        err=use_stderr,
+    )
+
+    spdx_identifiers = policy_config.get("spdx_identifiers", [])
+    if len(spdx_identifiers) > len(set(spdx_identifiers)):
+        raise click.BadParameter(
+            "SPDX Identifiers must be unique.", param="spdx_identifiers"
```
chrisimcevoy, line 252: Not anymore.

cloudsmith_cli/cli/tests/commands/policy/test_licence.py
```diff
@@ -0,0 +1,181 @@
...
+
+    data = {
+        "name": name,
+        "description": description,
+        "spdx_identifiers": list(spdx_identifiers),
+        "allow_unknown_licenses": allow_unknown_licenses,
+        "on_violation_quarantine": on_violation_quarantine,
+    }
+
+    file_path = directory / "License-POLICY-CONFIG.json"
```
chrisimcevoy, line 29: NAB: Could you make the casing consistent on the filename? Doesn't make a bit of difference, but... I'll know it's there.

cloudsmith_cli/cli/tests/commands/policy/test_licence.py
```diff
@@ -0,0 +1,181 @@
...
+    assert "Getting license policies ... OK" in result.output
+    # Ignoring the exact number of results and pluralization of the word "policy"
+    # here, because we may be in a system with pre-existing policies.
+    assert "Results:" in result.output
+    assert_output_matches_policy_config(result.output, policy_config_file_path)
+
+    # Change the values in the config file
+    policy_config_file_path = create_license_policy_config_file(
+        directory=tmp_path,
+        name=policy_name,
```
chrisimcevoy, line 136: If `policy_name` isn't used in the rest of this test, change this to `name=random_str()`

chrisimcevoy (approved):
Looks great, just a handful of tiny nitpicks. Approved in advance.

cloudsmith_cli/cli/commands/policy/license.py
Leeoc: TIL, will remove it now!

cloudsmith_cli/cli/commands/policy/license.py
Leeoc: They are required for the API, feel like an empty list defeats the point of creating the policy

chrisimcevoy: As in `spdx_identifiers = policy_config_file.get("spdx_identifiers", [])`. Maybe?

chrisimcevoy: Ah, but you may want to create a policy which only flags unknown licenses, no...? Anyway, not worth thinking about really. It's not that hard to specify an empty list in the config file and most people are probably going to follow the example config which includes that property.

Leeoc: Ahh thats a good point, will default it to an empty list cheers!

PR #116, opened 2023-05-11
Issue 115: Set null for name and email so whoami works on service acct

JesseRhoads-PD:
This should fix an issue we encountered where cloudsmith-cli can't do a whoami on a service account.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -21,3 +21,4 @@ We'd also like to say thank you to the following people (in alphabetical order),
 - Michal Nowak ([@Mno-hime](https://github.com/Mno-hime))
 - Rob Madole ([@robmadole](https://github.com/robmadole))
 - Sean Allen ([@SeanTAllen](https://github.com/SeanTAllen))
+- Jesse Rhoads ([@JesseRhoads-PD](https://github.com/JesseRhoads-PD))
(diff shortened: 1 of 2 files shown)


chrisimcevoy:
Hey Jesse 👋 
Thanks so much for taking the time to look at this.
It looks like what's happening here is that `email` is `None` for Services, which makes sense.
(We came across this a month or two ago, and found that this Exception wouldn't be raised if we moved to click v8.0, but that isn't possible until Python 2 support has been deprecated.)
One suggestion: Rather than outputting `"email: \"Null\""`, what would you say to _not_ outputting anything for `email` if it is falsy?
Happy to co-author something to that effect if you are ok with that?

JesseRhoads-PD:
Sure, feel free to take over and land this one differently, I just wanted to get a fix out ASAP. 
Feel free to Implement it in whatever way makes the most sense.

chrisimcevoy:
So now for a User, `cloudsmith whoami` output will be something like:

```
Retrieving your authentication status from the API ... OK
You are authenticated as:
Chris McEvoy (slug: chris-mcevoy, email: cmcevoy@cloudsmith.io)
```

And for a Service:

```
Retrieving your authentication status from the API ... OK
You are authenticated as:
bot-manager (slug: bot-manager)
```

Does that look ok to you, Jesse?

JesseRhoads-PD:
Yes, I put this in locally and it worked. Thanks!

chrisimcevoy (approved):

JesseRhoads-PD:
Rebased and signed so it can be merged.

PR #118, opened 2023-05-17
feat(eng-2374): Add policy package query string

Leeoc:
Adds `package_query_string` to the existing license and vulnerability policy CLI's

image


cloudsmith_cli/data/config.ini
```diff
@@ -20,4 +20,5 @@ api_user_agent=
 #  * -P your-profile-name (as an argument)
 #  * --profile your-profile-name (an an argument)
 #  * CLOUDSMITH_PROFILE=your-profile-name (as an env variable)
-[profile:your-profile-name]
+[profile:local]
+api_host=https://loconnell.cloudsmith.sh/api/v1
```
chrisimcevoy, line 24: Oops

requirements/development.py3.txt
```diff
@@ -1,6 +1,6 @@
 #
-# This file is autogenerated by pip-compile with python 3.6
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.11
```
Leeoc, line 2: Hmm should I recompile the requirement's using 3.6?

chrisimcevoy: Not sure. We're due to deprecate 2.7 in the 1.0 release, but other than that we support "Python 3". I don't foresee any issues installing those reqs with Python 3.6, but it couldn't hurt to check. FWIW I use 3.6 because it's the most recently-eol'ed CPython, and just to be sure to be sure.

Leeoc: Tried to regen using 3.6 (pyenv 3.6.15) think I am doing something wrong as it is the exact same. Even after deleting all existing requirement files and the `.venv`'s

chrisimcevoy:
I wonder do we need to start worrying about the length of some of these values when "pretty" printing.

Thinking if someone had a particularly cumbersome package_query_string, or a boatload of spdx_identifiers in a license policy?

For non-pretty output, we should show it all though.

cloudsmith_cli/cli/tests/commands/policy/test_vulnerability.py
chrisimcevoy, line 1: We need to assert that the package_query_string we get back from the api is what we expect.

cloudsmith_cli/cli/tests/commands/policy/test_licence.py
chrisimcevoy, line 1: We need to assert that the package_query_string we get back from the api is what we expect.

chrisimcevoy (changes requested):

chrisimcevoy (approved):
Nice one! LGTM.

PR #131, opened 2023-07-07
Eng 2493/upstream config cli

Leeoc:
Adds upstream configuration functionality.

### DIFF
--- a/cloudsmith_cli/cli/commands/upstream.py
+++ b/cloudsmith_cli/cli/commands/upstream.py
@@ -1,0 +1,472 @@
+# -*- coding: utf-8 -*-
+"""CLI/Commands - create, retrieve, update or delete repository upstreams."""
+from __future__ import absolute_import, print_function, unicode_literals
+
+import json
+
+import click
+import six
+
+from ...core.api import upstreams as api
+from .. import command, decorators, utils, validators
+from ..exceptions import handle_api_exceptions
+from ..utils import (
+    fmt_bool,
+    fmt_datetime,
+    maybe_spinner,
+    maybe_truncate_list,
+    maybe_truncate_string,
+)
+from .main import main
+
+UPSTREAM_FORMATS = [
+    "dart",
+    "deb",
+    "docker",
+    "helm",
+    "maven",
+    "nuget",
+    "npm",
+    "python",
+    "rpm",
+    "ruby",
+]
+
+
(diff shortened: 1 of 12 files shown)


cloudsmith_cli/cli/commands/upstream.py
```diff
@@ -0,0 +1,449 @@
...
+from ..utils import (
+    fmt_bool,
+    fmt_datetime,
+    maybe_spinner,
+    maybe_truncate_list,
+    maybe_truncate_string,
+)
+from .main import main
+
+UPSTREAM_FORMATS = [
```
MrTam, line 22: NAB: it's a shame we can't leverage the existing formats API to get a list of supported upstream formats. Perhaps enriching it with an attribute for `supports_upstream`, or query could be neat. This approach does work of course, but ties upstream format releases to a new CLI release.

For example, here we filter packages that use distros: https://github.com/cloudsmith-io/cloudsmith-cli/blob/5933eb1bf09f1d322d51f0112aae223e37116f1c/cloudsmith_cli/core/api/packages.py#L292-L294

cloudsmith_cli/cli/commands/upstream.py
```diff
@@ -0,0 +1,449 @@
...
+        "Extra Value 2",
+        "Active",
+        "Mode",
+        "Priority",
+        "Slug Perm",
+        "Updated At",
+        "Verify SSL",
+        "Component",
+        "Distribution Versions",
+        "Upstream Distribution",
```
MrTam, line 58: Similar to the above, reckon we could hide these for formats that don't support distros?

MrTam (approved):
LGTM, some suggestions to improve the 'dynamicity' (i don't think that's a word) of the format support and availability of parameters.

cloudsmith_cli/cli/commands/upstream.py
chrisimcevoy: True. Sounds like a ticket for QOL pile.

cloudsmith_cli/cli/commands/upstream.py
chrisimcevoy: Will do, cheers.

chrisimcevoy:
I was unable to merge as there were two commits from @Leeoc which were unsigned. I re-committed everything, should merge now.

PR #134, opened 2023-08-08
feat(eng-2346): Drop Python 2 support

chrisimcevoy:
- Drops support for python < 3.8. Going forward we will only support [python versions which are not EOL](https://endoflife.date/python).
- Reworked/simplified requirements declarations:
    - Production requirements are now declared in `setup.py`.
    - Dev requirements are now declared in `requirements.in`.
    - We still use pip-compile, but point it at `requirements.in` and `setup.py` to generate a `requirements.txt`, used for local dev and CI.
    - There are no interpreter/version-specific requirements files any more.
- Modernises the (versions of) pre-commit hooks we use.
- Uses a combination of direnv and pyenv to automagically set up dev env.
- direnv now sources a .env file to load the env vars required for pytest suite.
- Removes a bunch of other stuff (e.g. tox.ini) which either isn't used by everyone, or is just in the interest of simplification to get rid of.


MrTam (approved):
🚀 Sweet!

PR #136, opened 2023-08-15
chore: test against all supported runtimes

chrisimcevoy:
For python > 3.8, use circleci's python orb to pytest.

Existing 3.8-based stuff stays as-is, because the python orb doesn't integrate with code climate. (We can change that at a later stage if we want to use something other than code climate which the orb _does_ integrate with.)

Plus a couple of tweaks to tests to not be overly-strict about how many policies exist in an org, now that we're running multiple pytests in parallel. (This could be improved in future if we ever support creating an org from the cli...)

### DIFF
--- a/bin/circle-exec
+++ b/bin/circle-exec
@@ -13,4 +13,4 @@ trap cleanup EXIT
 circleci config process .circleci/config.yml > $WORK_DIR/config.yml
 cat $WORK_DIR/config.yml
 circleci local execute --config $WORK_DIR/config.yml \
-  --env "CIRCLE_LOCAL=true"
+  --env "CIRCLE_LOCAL=true" "$@"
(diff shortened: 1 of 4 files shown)


bin/circle-exec
```diff
@@ -13,4 +13,4 @@ trap cleanup EXIT
 circleci config process .circleci/config.yml > $WORK_DIR/config.yml
 cat $WORK_DIR/config.yml
 circleci local execute --config $WORK_DIR/config.yml \
-  --env "CIRCLE_LOCAL=true"
+  --env "CIRCLE_LOCAL=true" "$@"
```
dmeecs, line 16: What's the @ thing here for? Just wondering.

chrisimcevoy: It's for passing the job name to circleci cli, so you can use this script to run the jobs locally. (accidentally removed in a previous pr)

PR #140, opened 2023-08-28
chore: tests for api init and rest client retry

chrisimcevoy:
Added tests for `initialise_api()` and retry functionality in `RestClient.request()`.

I really just wanted to test the retry stuff, but you can't instantiate a rest client without doing the api initialisation first.

Side note:
I found that the more popular requests mocking libraries( e.g. [responses](https://github.com/getsentry/responses/issues/135) and [requests-mock](https://github.com/jamielennox/requests-mock/issues/83)) conflict with retry functionality in our requests adapter, because they too operate (i.e. mock) at the requests adapter level. httpretty operates at the socket level, which makes it possible to test our gear.

### DIFF
--- a/cloudsmith_cli/core/rest.py
+++ b/cloudsmith_cli/core/rest.py
@@ -12,7 +12,7 @@ import requests.exceptions
 from cloudsmith_api.configuration import Configuration
 from cloudsmith_api.rest import ApiException, RESTClientObject
 from requests.adapters import HTTPAdapter
-from urllib3.util.retry import Retry  # pylint: disable=import-error
+from urllib3.util.retry import Retry
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


cloudsmith_cli/core/rest.py
```diff
@@ -12,7 +12,7 @@
 from cloudsmith_api.configuration import Configuration
 from cloudsmith_api.rest import ApiException, RESTClientObject
 from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util.retry import Retry  # pylint: disable=import-error
+from urllib3.util.retry import Retry  # pylint: disable=import-error
```
jackgibson1, line 15: Just curious - why is pylint complaining about an import error here?

chrisimcevoy: Good catch. Changed the import but forgot to remove the pylint-disable. Removed it now.

jackgibson1 (approved):

PR #160, opened 2024-03-20
Update config.py

keithy:
Correction to error in the schema


dkatavic (approved):
Looks good @keithy , I'll take a look at that failing test

PR #163, opened 2024-04-10
ceng-163: Fix repository pagination

BartoszBlizniak:
Before the change, the repository list command didn't return pagination: 
image

After the change, the pagination is back:
image


alancarson (approved):
Looks great. Thank you for fixing this.

PR #164, opened 2024-06-11
CENG-260: update-cloudsmith-cli-to-include-zipapp-in-the-release

BartoszBlizniak:
* Added zipapp of the cloudsmith-cli
* Added workflow to generate the release with `.pyz` file

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,45 @@
+name: Create Zipapp and Release
+
+on:
+    push:
+        tags:
+            - "v*"
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - name: Set up Python 3.9
+              uses: actions/setup-python@v2
+              with:
+                  python-version: 3.9
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip
+                  pip install shiv
+            - name: Get version
+              id: get_version
+              run: echo "VERSION=$(cat cloudsmith_cli/data/VERSION)" >> $GITHUB_ENV
+            - name: Create Zipapp
+              run: shiv -o cloudsmith-${{ env.VERSION }}.pyz -c cloudsmith .
+            - name: Create Release
+              id: create_release
+              uses: actions/create-release@v1
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+              with:
+                  python-version: 3.9
+
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip
+                  pip install shiv
+
+            - name: Create Zipapp
+              run: shiv -o cloudsmith-cli.pyz -e cloudsmith_cli.cli.commands.main:main .
```
paddycarey, line 26: do we need to manually specify the entrypoint here? can we just `-c cloudsmith` instead?

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+
+            - name: Upload Release Asset
+              id: upload-release-asset
+              uses: actions/upload-release-asset@v1
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              with:
+                  upload_url: ${{ steps.create_release.outputs.upload_url }}
+                  asset_path: ./cloudsmith-cli.pyz
+                  asset_name: cloudsmith-cli.pyz
```
paddycarey, line 48: should we put the version number in the name? that's usually best practice, just makes it simpler for people downloading multiple versions

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
+name: Create Zipapp and Release
```
paddycarey, line 1: Should we do this in CircleCI since that's where the rest of the CI for this repo already runs?

.github/workflows/release.yml
paddycarey: When you `pip install` the CLI it uses the name `cloudsmith`, I think we should match that, rather than `cloudsmith-cli`

paddycarey (commented):
Looks like it does what it says on the tin! Few questions just.

.github/workflows/release.yml
BartoszBlizniak: I'm not that familiar with CircleCi so decided to do it in GHA instead. I did briefly take a look at it but GHA has a nicer and easier way of implementing this as part of their built-in action and uses pre-defined credentials. If this is a biggie, I can take a look at it but if you're happy to roll with this and I can take a proper look at this when I get some more time.

paddycarey (approved):

PR #168, opened 2024-08-28
NO-TICKET: Upload zipapp to CS

BartoszBlizniak:
* Uploads the zipapp to Cloudsmith repository


.github/workflows/release.yml
```diff
@@ -18,6 +18,7 @@ jobs:
               run: |
                   python -m pip install --upgrade pip
                   pip install shiv
+                  pip install cloudsmith-cli
```
paddycarey, line 21: do you need to install this? you're building the CLI so can you use the one you've just built to push itself?

Issue #170, opened 2024-09-20
Add flag to get all pages


tlimoncelli:
This utility would be more useful if "list repos" and "list pkgs" could get all pages.  Scripting with this utility would be considerably easier as implementing the algorithm to get all pages is rather difficult in bash.

nickbroon:
This would significantly enhance the utility and ergonomics of the cloudsmith CLI.

nickbroon:
> Looks like [#172](https://github.com/cloudsmith-io/cloudsmith-cli/pull/172) added this feature. Thanks!

However that appears to not have been merged.

PR #171, opened 2024-09-20
Get all pages

TomOnTime:
Fixes https://github.com/cloudsmith-io/cloudsmith-cli/issues/170

This is a work in progress.

This codes enables `ls repos` and `ls pkgs` to get all pages.

I wanted to add a flag `--all-pages` but I couldn't figure out how. I made a couple attempts to add the appropriate click statements but I couldn't get the value to propagate to where it was needed.  I could use some help!

In the meanwhile I added a (brutal but temporarily) way to activate this code.  The (undocumented) `--page -1` enables this feature.

```
$ python3 -m cloudsmith_cli ls pkgs --page-size 3 --page -1  stackoverflow/raw
Getting list of packages ... OK
...
...
[REDACTED]                  | 2023.1.17          | Completed | [REDACTED]/raw/[REDACTED]               

Results: 62 packages visible
```

Could you help me add a proper flag?  Ideally this should be the default, with the `--page x` enabling the old behavior for backwards compatibility.

### DIFF
--- a/cloudsmith_cli/cli/commands/list_.py
+++ b/cloudsmith_cli/cli/commands/list_.py
@@ -174,11 +174,26 @@ def packages(ctx, opts, owner_repo, page, page_size, query):
     click.echo("Getting list of packages ... ", nl=False, err=use_stderr)
 
     context_msg = "Failed to get list of packages!"
-    with handle_api_exceptions(ctx, opts=opts, context_msg=context_msg):
-        with maybe_spinner(opts):
-            packages_, page_info = list_packages(
-                owner=owner, repo=repo, page=page, page_size=page_size, query=query
-            )
+    if page != -1:
+        with handle_api_exceptions(ctx, opts=opts, context_msg=context_msg):
+            with maybe_spinner(opts):
+                packages_, page_info = list_packages(
+                    owner=owner, repo=repo, page=page, page_size=page_size, query=query
+                )
+    else:
+        with handle_api_exceptions(ctx, opts=opts, context_msg=context_msg):
+            with maybe_spinner(opts):
+                packages_ = []
+                page_number = 1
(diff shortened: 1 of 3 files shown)


tlimoncelli:
After thinking about it a bit, let me suggest this instead:

The default should be to get all pages.  I can't imagine a situation where someone would intentionally want partial information.  In that case, my suggestion is to simply have `ls pkgs` and `ls repos` ignore the `--page` and `--page-size` flags and always return the complete information.

However a case can be made for backwards compatibility. There may be scripts that issue `--page 1`, `--page 2`, etc. and would break.  On one hand, we can simply let those scripts break as the authors would probably be very happy to not have to assemble the individual pages any more.  On the other hand, you could offer backwards compatibility if the `--page` flag is used. That is, "all pages" would be the default but any use of the `--page` flag would trigger backwards compatibility.

Neither would require any new flags.

I'd be glad to implement the first suggestion. I'd need assistance implementing the second.

tlimoncelli:
Hi @BartoszBlizniak 

Sounds good.

I made a few attempts at adding a flag with no success.  Can you assist?   I tried adding the flag to `common_cli_list_options` in `cloudsmith_cli/cli/decorators.py` but I can't see how to get the value to propagate to `cloudsmith_cli/cli/commands/list_.py`

BartoszBlizniak:
> Hi @BartoszBlizniak
> 
> Sounds good.
> 
> I made a few attempts at adding a flag with no success. Can you assist? I tried adding the flag to `common_cli_list_options` in `cloudsmith_cli/cli/decorators.py` but I can't see how to get the value to propagate to `cloudsmith_cli/cli/commands/list_.py`

No problem, I will unfortunately have to create a branch myself due to limitations for tests running from Fork. Once I have everything ready I will reference and credit the contribution!

PR #174, opened 2024-10-07
Add support for SAML authentication

AndrewSpeed:
# What changed?

Adds the `cloudsmith auth` command to authenticate using an organization's configured SAML provider rather than an API key

Access and refresh tokens are stored in the operating system keyring with the timestamp they were generated at. On each command, if the access token is set, the CLI will check if the token should be refreshed prior to executing the command specified

As part of the refresh, the timestamp denoting when the last refresh was attempted will be updated, ensuring we don't repeatedly try to refresh the token on multiple invocations if the refresh failed

### DIFF
--- a/cloudsmith_cli/cli/commands/auth.py
+++ b/cloudsmith_cli/cli/commands/auth.py
@@ -37,7 +37,7 @@ def authenticate(ctx, opts, owner):
     )
     org_saml_url += urlencode({"redirect_url": "http://localhost:12400"})
 
-    org_saml_response = requests.get(org_saml_url, timeout=5)
+    org_saml_response = requests.get(org_saml_url, timeout=30)
     idp_url = org_saml_response.json().get("redirect_url")
 
     click.echo(
@@ -55,5 +55,7 @@ def authenticate(ctx, opts, owner):
     )
     auth_server.handle_request()
 
+    click.echo()
     click.echo(f"Access token: {get_access_token()}")
+    click.echo()
     click.echo(f"Refresh token: {get_refresh_token()}")
--- a/cloudsmith_cli/cli/commands/auth.py
+++ b/cloudsmith_cli/cli/commands/auth.py
@@ -26,7 +26,6 @@ from .main import main
 @click.pass_context
 def authenticate(ctx, opts, owner):
     """Authenticate to Cloudsmith using the org's SAML setup."""
-    # TODO: Why is a single arg a list?
     owner = owner[0]
 
     click.echo(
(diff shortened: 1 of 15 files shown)


cloudsmith_cli/cli/commands/auth.py
```diff
@@ -0,0 +1,57 @@
...
+        click.echo(
+            "Opening your organization's SAML IDP URL in your browser: %(idp_url)s"
+            % {"idp_url": click.style(idp_url, bold=True)}
+        )
+        click.echo()
+        webbrowser.open(idp_url)
+        click.echo("Starting webserver to begin authentication ... ")
+
+        auth_server = AuthenticationWebServer(
+            ("0.0.0.0", 12400),
```
cloudsmith-iduffy, line 49: Should we use 127.0.0.1 here so that the web server isn't exposed to everything on the network?

AndrewSpeed: 💡 Yes, we absolutely should!

cloudsmith-iduffy (approved):
Installed locally and successfully authenticated using SSO.

cloudsmith_cli/core/api/init.py
```diff
@@ -40,15 +44,51 @@ def initialise_api(
...
+                    )
+                    keyring.store_sso_tokens(
+                        config.host, new_access_token, new_refresh_token
+                    )
+            except ApiException:
+                keyring.update_refresh_attempted_at(config.host)
+
+                # try falling back to API key auth if refresh fails
+                if key:
+                    config.api_key["X-Api-Key"] = key
```
estebangarcia, line 68: ❓ Should we communicate back to the user here that their session might've expired? The AWS CLI returns something like
```CredentialsProviderError: The SSO session associated with this profile has expired. To refresh this SSO session run aws sso login with the corresponding profile```

Ignore if a similar message is returned already from another part of the code.

File: cloudsmith_cli/cli/commands/auth.py [hunk 3]
```diff
@@ -0,0 +1,58 @@
...
+def authenticate(ctx, opts, owner):
+    """Authenticate to Cloudsmith using the org's SAML setup."""
+    owner = owner[0]
+    api_host = opts.api_config.host
+
+    click.echo(
+        "Beginning authentication for the {owner} org ... ".format(
+            owner=click.style(owner, bold=True)
+        )
+    )
```
estebangarcia, line 36: NAB: Would be nice if we allow for configuring the owner in a profile instead so we follow the same practice we already have and it aligns with how AWS works too.

estebangarcia (approved):
Awesome work 🎉🎉🎉

cloudsmith_cli/cli/commands/auth.py
AndrewSpeed: > if we allow for configuring the owner in a profile instead

Do you mean in the CLI profile configuration, or some other mechanism?

PR #183, opened 2024-12-20
CENG-450: Fix `cloudsmith auth` problem when user passed owner

BartoszBlizniak:
Fix:
image


cloudsmith_cli/data/VERSION
```diff
@@ -1 +1 @@
-1.4.1
+1.4.2
```
chrisimcevoy, line 1: Please use `bumpversion` (in a separate PR) rather than changing this file by hand.

BartoszBlizniak: done ✅

cloudsmith-iduffy (approved):

PR #189, opened 2025-03-22
feat: add HTML templates for authentication success and error, and improve session handling in sso

cloudsmith-iduffy:
- Creating a session in the SSO/SAML flow so that similar headers/configuration that are passed to the API client are also used for the http client that handles the saml authentication
 - Add support for JSON formatting on the whoami response
 - Add HTML templates for the SSO/SAML callback
 - Add support for parsing ?error query params on the SSO/SAML callback
 - Switch to UV
Screenshot 2025-03-22 at 22 41 15
Screenshot 2025-03-22 at 22 41 05


cloudsmith_cli/cli/webserver.py
```diff
@@ -9,19 +12,52 @@
...
+        Rendered HTML content
+    """
+    template_path = get_template_path(template_name)
+
+    with open(template_path, encoding="utf-8") as file:
+        content = file.read()
+
+    # Replace placeholders with context values
+    for key, value in context.items():
+        placeholder = f"<!-- {key.upper()}_PLACEHOLDER -->"
```
dgonzalez, line 39: suggestion: You can use [Jinja2 templates](https://jinja.palletsprojects.com/en/stable/templates/) here. The current approach works but with Jinja2, it is simpler as you set the context and the variables are available in the template.

dgonzalez (approved):
Two small suggestions but it is good as it is.

cloudsmith_cli/cli/webserver.py
cloudsmith-iduffy: I agree, but I didn't want to introduce a new dependency for such a small use case. Happy to adjust if you really want it.

PR #192, opened 2025-04-08
Support for managing User API Tokens via SAML authentication

LowzG:
For environments that only use SAML for authentication, there currently is no way to programmatically retrieve an existing User API Token. Currently the Cloudsmith CLI allows creating/retrieving User API Tokens only via the `login/token` command which uses basic auth. 

This is my hacky way of getting around these limitations. The `auth --token` command brings consistency between the `login` and `auth` commands, with one caveat: users with an existing token will be asked if they would like to refresh their token.

This also includes the `tokens` command that allows users to list and refresh their token.

To remove the hacky parts of this implementations we need:

1. An API endpoint that allows for the retrieval of an existing token that supports the `Authorization` header without basic auth. 
2. A class [here](https://github.com/cloudsmith-io/cloudsmith-api/tree/master/bindings/python/src/cloudsmith_api/api) that covers the endpoint(s) described in 1

### DIFF
--- a/cloudsmith_cli/cli/commands/tokens.py
+++ b/cloudsmith_cli/cli/commands/tokens.py
@@ -38,9 +38,9 @@ def list_tokens(ctx, opts):
 
     for token in tokens:
         click.echo(
-            f"Token: {click.style(token['key'], fg='magenta')}, "
-            f"Created: {click.style(token['created'], fg='green')}, "
-            f"slug_perm: {click.style(token['slug_perm'], fg='cyan')}"
+            f"Token: {click.style(token.key, fg='magenta')}, "
+            f"Created: {click.style(token.created, fg='green')}, "
+            f"slug_perm: {click.style(token.slug_perm, fg='cyan')}"
         )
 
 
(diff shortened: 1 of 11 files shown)


cloudsmith_cli/core/api/user.py
```diff
@@ -28,6 +28,22 @@ def get_user_token(login, password):
     return data.token
 
 
+def create_user_token_saml() -> dict:
+    """Create a new user API token using SAML."""
+    client = get_api_client(BaseApi)
+
+    with catch_raise_api_exception():
+        data, _, headers = client.api_client.call_api(
```
cloudsmith-iduffy, line 36: FYI: that endpoint isn't yet public/documented https://api.cloudsmith.io/swagger/ the team is working to change this very soon.

nickxn:
Thanks @LowzG - we're reviewing at the moment!

cloudsmith_cli/core/api/init.py
```diff
@@ -12,6 +12,14 @@
 from .exceptions import ApiException
 
 
+class BaseApi:
```
LowzG, line 15: Maybe we should remove this class. Now that we can use `get_user_api()` it is no longer needed.

mhutchings-182: Good shout! Thanks!

cloudsmith_cli/cli/commands/tokens.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if maybe_print_as_json(opts, tokens):
+        return
+
+    for token in tokens:
+        click.echo(
+            f"Token: {click.style(token.key, fg='magenta')}, "
+            f"Created: {click.style(token.created, fg='green')}, "
+            f"slug_perm: {click.style(token.slug_perm, fg='cyan')}"
+        )
```
jmccayCS, line 44: NAB: could maybe create a helper function to print this output? Seen it used in a few places

jmccayCS (approved):
LGTM just one small comment there but not a blocker

PR #204, opened 2025-05-26
Bump version: 1.8.1 → 1.8.2

cloudsmith-iduffy:


chrisimcevoy (approved):
Anything to add to the changelog here?

dmeecs:
> Nothing to add, just releasing the sdist alongside the wheel.
> 
> Version bump only to generate a new release without causing conflicts.

IMO @cloudsmith-iduffy that's something worth putting in the changelog otherwise people will think what's in version 1.8.2.

PR #218, opened 2025-10-16
Ceng 567 fix sso auth token refresh workflow

colinmoynes:
This PR addresses a workflow issue for being able to programmatically refresh tokens via the CLI. Previously, this required interactive prompts which broke automated pipelines. 

* adds —-force parameter to both the 'auth' and 'tokens refresh' commands, which circumvents the interactive prompts and automatically refreshes the token. 

This was done as part of CENG-567. 

Tested internally.


cloudsmith_cli/cli/commands/auth.py
```diff
@@ -31,11 +31,18 @@
     is_flag=True,
     help="Retrieve a user API token after successful authentication.",
 )
+@click.option(
+    "-f",
+    "--force",
+    default=False,
+    is_flag=True,
+    help="Force create a user API token after successful authentication.",
```
paddycarey: does this accurately describe what's happening? from reading the code it looks like this is actually a forced rotation, rather than creation?

colinmoynes: You are correct. Missed updating this since the original idea was to create, but that's not the case now.

PR #244, opened 2025-12-16
feat: MCP Server

estebangarcia:
# Add MCP Server Support to Cloudsmith CLI

This PR introduces [Model Context Protocol (MCP)](https://modelcontextprotocol.io/) server support to the Cloudsmith CLI, enabling AI assistants and other MCP clients to interact with Cloudsmith's API programmatically.

## What is MCP?

The Model Context Protocol is an open standard that enables AI assistants to securely connect to external data sources and tools. By adding MCP server support to the Cloudsmith CLI, users can now leverage AI assistants (like Claude Desktop, and others) to manage their Cloudsmith repositories, packages, and artifacts through natural language.

## Key Features

* Dynamic Tool Generation: MCP tools are automatically generated from Cloudsmith's OpenAPI specification, ensuring the server stays in sync with the API
* Optimized API Responses: JSON responses are encoded to TOON format and use sparse fieldsets (taken from the `x-simplified` parameter in the OpenAPI spec) to minimize token usage and improve performance
* Configurable Tool Exposure: Control which tools and tool groups are exposed to prevent context window overflow in MCP clients
* Seamless Authentication: Authentication is handled by the CLI using existing Cloudsmith API keys or SSO credentials
* Multi-Profile Support: Configure separate MCP server instances for different Cloudsmith profiles

## Default Available Tools

The MCP server dynamically generates tools from Cloudsmith's OpenAPI specification, which results in a very large number of available tools. While this provides complete API coverage, exposing all tools simultaneously would immediately fill the LLM's context windows.
To address this, the server exposes only a curated subset of commonly-used tools by default. Users can customize which tools and tool groups are available based on their specific workflows, ensuring MCP clients remain efficient and responsive while still providing access to the full API when needed.

The list of tools is filtered by disabling certain categories found here https://github.com/cloudsmith-io/cloudsmith-cli/blob/eng-9528/mcp-integration/cloudsmith_cli/core/mcp/server.py#L38

## Commands Added

```bash
# Start the MCP server
$ cloudsmith mcp start

# List available tools (use -a to show all tools)
$ cloudsmith mcp list_tools

# List available tool groups (use -a to show all groups)
$ cloudsmith mcp list_groups

# Auto-configure MCP clients (use -P to configure for a specific profile)
$ cloudsmith mcp configure
```

## Configuration

Control which tools are exposed by adding configuration to `~/.cloudsmith/config.ini`:
```
mcp_allowed_tools=workspaces_policies_simulate_list
mcp_allowed_tool_groups=metrics
```

This exposes the specified individual tools and all tools within the listed tool groups.

## Breaking Changes

**This release requires Python 3.10 or later due to MCP SDK dependencies.**

## Additional Notes

* Authentication uses the existing CLI credentials (API keys or SSO)
* For SSO authentication, the auth flow must be completed via `cloudsmith auth` before starting the MCP server as MCP clients will not trigger the SSO authentication flow automatically


cloudsmith_cli/cli/commands/mcp.py
```diff
@@ -0,0 +1,355 @@
...
+    # Ensure parent directory exists
+    config_path.parent.mkdir(parents=True, exist_ok=True)
+
+    # Read existing config or create new one
+    if config_path.exists():
+        with open(config_path) as f:
+            try:
+                config = json.load(f)
+            except json.JSONDecodeError:
+                config = {}
```
apoclyps: *thoughts* It's an edge case, but if the configuration is invalid (e.g., due to a typo/user editing that prevents decoding), wouldn’t setting the config to `{}` result in the configuration being overridden/replaced? Should we instead raise an exception here to call out that the config is wrong? 

Happy to go either way, as it depends on whether we think users will modify it or not to do the least surprising thing in that scenario.

apoclyps (approved):

cloudsmith_cli/cli/commands/mcp.py
estebangarcia: Good point. I change it to raise an error.

cloudsmith_cli/cli/commands/mcp.py
```diff
@@ -0,0 +1,355 @@
...
+import sys
+from pathlib import Path
+
+import click
+
+from ...core.mcp import server
+from .. import command, decorators, utils
+from .main import main
+
+SUPPORTED_MCP_CLIENTS = ["claude", "cursor", "vscode"]
```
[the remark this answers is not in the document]
estebangarcia: Yeah I included the most used ones, I added Gemini CLI too now.

We can look at it again on the next version of it.

PR #248, opened 2025-12-17
fix(ceng-630): Fix JSON output across all commands

BartoszBlizniak:
This pull request significantly improves the Cloudsmith CLI's handling of output when the `--output-format json` option is used. It ensures that all informational messages, warnings, and prompts are sent to stderr, while error messages from command results are properly formatted as structured JSON on stdout. The changes will also output click exceptions in JSON format to ensure consistent results being returned when things fail.

The most important changes are:

**Deprecation warning:**

* The `--json` parameter used as part of the `cloudsmith auth` command will be removed in a future release of the CLI. Users are asked to switch back to using `--output-format json` starting from this release.

**Output Handling and JSON Support:**

* All commands now route informational messages, warnings, and interactive prompts to stderr when `--output-format json` is active, and output error messages as structured JSON on stdout. This ensures clean and machine-readable JSON output for all commands.
* Introduced and standardized the use of the `utils.should_use_stderr(opts)` utility function across all commands to determine when to send output to stderr, replacing previous ad-hoc checks. 

**Command Updates:**

* Updated commands such as `auth`, `check`, `copy`, `delete`, `download`, `entitlements`, `list_`, and `login` to use the new stderr logic for informational and prompt messages, and to output results/errors as JSON when requested. 

**Examples:**

Shell redirect will hide all output and present only the result in JSON:

```
> cloudsmith push raw bart-demo-org/618 README.md -F json --republish --no-wait-for-sync 2>/dev/null | jq
{
  "data": {
    "slug": "readmemd-wzmo",
    "slug_perm": "479DjKoMYEHd",
    "status": "OK"
  }
}
```

Example download command:

```
> cloudsmith download bart-demo-org/618 common --outfile pom.xml -F json | jq
{
  "data": {
    "filename": "pom.xml",
    "package": {
      "format": "maven",
      "name": "common",
      "slug": "common-03pom-n6gy",
      "version": "0.3"
    },
    "path": "/Users/**redacted**/Downloads/pom.xml",
    "size": 200913,
    "status": "OK"
  }
}
```

Error example (Note the information message is coming from stderr so the JSON is left intact and parasable)

```
> cloudsmith auth --token -o bart-demo-org -F json | jq
Beginning authentication for the bart-demo-org org ... 
{
  "detail": "Not Found",
  "help": {
    "context": "Failed to authenticate via SSO!",
    "hint": "This usually means the user/org is wrong or not visible."
  },
  "meta": {
    "code": 404,
    "description": "Not Found"
  }
}
```
Click error example:

```
> cloudsmith list packages bart-demo-org/816 -F json --page 1 --page-all | jq
{
  "detail": "Invalid value for '--page-all' / '--show-all': Cannot be used with --page (-p) or --page-size (-l). (--show-all is an alias for --page-all)",
  "meta": {
    "code": 2,
    "description": "Usage Error"
  },
  "help": {
    "context": "Invalid usage",
    "hint": "Check your command arguments/flags."
  }
}
```

**JSON Error Schema:**

```
{
  "detail": "string",
  "fields": {
    "field_name": ["string (error message)", "string (error message)"]
  },
  "help": {
    "context": "string",
    "hint": "string (optional)"
  },
  "meta": {
    "code": 400,
    "description": "string (e.g. Bad Request)"
  }
}
```


cloudsmith_cli/cli/commands/download.py
```diff
@@ -321,12 +374,28 @@ def download(  # noqa: C901
...
+            "name": package.get("name"),
+            "version": package.get("version"),
+            "format": package.get("format"),
+            "slug": package.get("slug"),
+        },
+        "filename": os.path.basename(outfile),
+        "path": outfile,
+        "size": package.get("size", 0),
+        "status": "OK",
+    }
```
estebangarcia: nitpick/question: I'm thinking this JSON would be slightly different than the multi-file one I see above. If people are using jq filters they might expect outputs to be the same.

estebangarcia (approved):

cloudsmith_cli/cli/commands/download.py
BartoszBlizniak: Thanks, Esteban - I've aligned it to be similar output to the multi-file one by grouping it under the files array (in this case, it will always be one file) to standardise the output.

cloudsmith_cli/cli/command.py
```diff
@@ -6,6 +6,46 @@
...
+
+    if "--output-format=json" in argv or "--output-format=pretty_json" in argv:
+        return True
+
+    for idx, arg in enumerate(argv):
+        if arg in ("-F", "--output-format") and idx + 1 < len(argv):
+            if argv[idx + 1] in ("json", "pretty_json"):
+                return True
+
+    return False
```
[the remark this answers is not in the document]
BartoszBlizniak: Hey @paddycarey  just checking if we're okay to resolve this or if there is another approah I should take here?

PR #249, opened 2025-12-18
Add --tag option to download command

opswithranjan:
### Summary
This PR adds tag filtering capability to the `cloudsmith download` command and updates the README with comprehensive download command documentation.

#### 🚀 New Feature: --tag Option
- Added `--tag` option to `cloudsmith download` command for filtering packages by tags
- Enhanced `resolve_package()` function to support tag filtering across multiple metadata fields:
  - Actual package tags (info, version categories)
  - Package format (deb, rpm, python, etc.)
  - Architecture (arm64, amd64, etc.)
  - Distribution name/version (ubuntu, noble, ubuntu/noble)
  - Components (main, contrib, etc.)
- Added comprehensive unit tests for tag filtering functionality

#### 📚 Documentation Updates
- Added `download` command to README features list
- Created new "Downloading Packages" section with practical examples
- Documented all filtering options including the new `--tag` feature
- Provided examples for common use cases (version, format, arch, tag filtering)

### Usage Examples
```bash
# Filter by tag
cloudsmith download org/repo package --tag latest

# Combine with other filters  
cloudsmith download org/repo package --tag stable --format deb --arch arm64

# Preview downloads
cloudsmith download org/repo package --tag beta --dry-run
```

### Testing
- All existing tests pass
- New unit test `test_resolve_package_with_tag_filter` validates tag filtering
- Syntax validation and linting compliance verified
- Manual testing confirmed tag filtering works with real API responses

### Breaking Changes
None - this is a backward-compatible feature addition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added `--tag` option to `download` command for filtering packages by tags
+- Added download command documentation to README with comprehensive usage examples
 
 ## [1.13.0] - 2026-02-16
 
 ### Added
 
-- Added `--tag` option to `download` command for filtering packages by tags
-- Enhanced tag filtering to support all metadata fields shown as tags in UI (format, architecture, distribution, component, etc.)
-- Added download command documentation to README with comprehensive usage examples
 - Added `CLOUDSMITH_NO_KEYRING` environment variable to disable keyring usage globally. Set `CLOUDSMITH_NO_KEYRING=1` to skip system keyring operations.
 - Added `--request-api-key` flag to `cloudsmith auth` command for fully automated, non-interactive API token retrieval. Auto-creates a token if none exists, or auto-rotates (with warning) if one already exists. Compatible with `--save-config` and `CLOUDSMITH_NO_KEYRING`.
 - Added `--verbose` (`-v`) flag to `cloudsmith whoami` to show detailed authentication information including active method (API Key or SSO Token), credential source, token metadata, and SSO status. Supports `--output-format json`.


cloudsmith_cli/core/download.py
```diff
@@ -53,6 +53,50 @@ def resolve_auth(
...
+    Returns:
+        True if package matches the tag filter
+    """
+    # Check actual tags field (info, version, etc.)
+    pkg_tags = pkg.get("tags", {})
+    for tag_category in pkg_tags.values():
+        if isinstance(tag_category, list) and tag_filter in tag_category:
+            return True
+
+    # Check other metadata fields that appear as tags in the UI (case-sensitive)
```
BartoszBlizniak: Might be worthwhile double-checking if Cloudsmith Tags are actually case-sensitive in the API

BartoszBlizniak (commented):
Overall looks good - Got a few questions:

1. I can see we're starting to use a lot of "filters" for the download command. All of these are essentially being passed to the query parameter for listing packages. From code maintenance and usability, perhaps adding `package_query` instead of filters could allow for a better user experience (see package list command). -> This would require some re-working as it would be a breaking change, removing the filters (Future PR).
2. Integration test for the tag appears to be missing (you have core tests ✅)
3. The `--tag`, `--os` and `--arch` appear to be case sensitive, while `--name` isn't, should this be the case (no pun intended 😄)?

opswithranjan:
Thanks @BartoszBlizniak 
All valid points!

1. Agreed! Using package_query would be cleaner. Would be happy to tackle a migration PR that adds package_query while keeping filters as convenience aliases.
2. You're right - I will add an integration test for the tag filtering. 
3. Intentional choice: `--name` is case-insensitive for search convenience, but `--tag/--os/--arch` are case-sensitive to match exact API metadata values. For example, API returns "arm64" not "ARM64", and "Ubuntu" not "ubuntu". This ensures users specify tags exactly as they appear in the API/UI. Open to making them case-insensitive if preferred for better UX, though!

estebangarcia (approved):

PR #276, opened 2026-03-13
feat: add OIDC credential auto-discovery

cloudsmith-iduffy:
Add a new OIDC credential provider to the credential chain that
automatically detects CI/CD environments, retrieves a vendor OIDC JWT,
and exchanges it for a short-lived Cloudsmith API token.

Key changes:
- New OidcProvider in the credential provider chain (lowest priority)
- AWS environment detector using boto3/STS (optional `aws` extra)
- OIDC token exchange
- Token caching via system keyring with filesystem fallback
- `whoami` command updated to display OIDC auth source
- README updated with optional dependency install instructions

PR #300, opened 2026-06-08
feat: add GitHub Actions OIDC detector

cloudsmith-iduffy:
## Summary

Adds a **GitHub Actions** environment detector to the OIDC credential auto-discovery chain, mirroring the existing AWS detector (#276). When running in GitHub Actions with `id-token: write` permission, the CLI fetches an OIDC JWT from the Actions runtime HTTP endpoint and exchanges it for a short-lived Cloudsmith API token.

- Detects via `GITHUB_ACTIONS` + `ACTIONS_ID_TOKEN_REQUEST_URL`/`ACTIONS_ID_TOKEN_REQUEST_TOKEN`
- Reuses the shared `requests` session and audience conventions from the AWS detector (`self.context.oidc_audience or "cloudsmith"`)
- Ordered **before** the AWS detector so a GitHub runner with incidental AWS credentials still authenticates as GitHub Actions
- Needs **no extra dependencies** (plain HTTP) — unlike the AWS extra (`boto3`)

## Testing

Unit tests cover detection (all env permutations), token extraction, URL/audience/header construction, custom audience, the `&` separator when the request URL already has a query, and the missing-token error.

Validated in a real GitHub Actions pipeline — see the comment below for the live run.

References:
- https://docs.github.com/en/actions/reference/security/oidc
- https://docs.cloudsmith.com/authentication/setup-cloudsmith-to-authenticate-with-oidc-in-github-actions


cloudsmith-iduffy:
### Verified in a real GitHub Actions pipeline ✅

Ran the detector end-to-end in a throwaway repo ([cloudsmith-iduffy/cloudsmith-cli-oidc-test](https://github.com/cloudsmith-iduffy/cloudsmith-cli-oidc-test)) that installs the CLI from this branch and runs OIDC auto-discovery against `iduffy-demo` / `github-09kg`:

**Run:** https://github.com/cloudsmith-iduffy/cloudsmith-cli-oidc-test/actions/runs/27146690255/job/80126926128

The new detector works as intended:

- ✅ **Detection + token fetch succeed** — the GitHub Actions detector fires and successfully fetches the OIDC JWT from the Actions runtime endpoint. We never see `Failed to retrieve identity token from GitHub Actions`, and the exchange step only runs *after* `get_token()` returns a token.

The only error in the run is on the **Cloudsmith side**, not in this code:

```
OIDC: Token exchange failed: OIDC token exchange failed with 401: Unable to retrieve providers.
```

That `401: Unable to retrieve providers` means the `github-09kg` service account in `iduffy-demo` doesn't yet have an OIDC provider configured for GitHub's issuer (`https://token.actions.githubusercontent.com`) with the right audience/subject claims — a config prerequisite, independent of the detector. Once that trust policy is in place, the same workflow completes the full round-trip and `whoami --verbose` reports `Source: OIDC via GitHub Actions`.

BartoszBlizniak (approved):

PR #304, opened 2026-06-08
feat: add generic OIDC detector

cloudsmith-iduffy:
## What

Adds a **generic fallback OIDC detector** that reads an OIDC JWT from the `CLOUDSMITH_OIDC_TOKEN` environment variable.

This supports:
- **Jenkins** (via the [credentials binding plugin](https://plugins.jenkins.io/credentials-binding/))
- Any custom CI/CD system that can inject an OIDC token via an environment variable

## How

- New `GenericDetector` in `cloudsmith_cli/core/credentials/oidc/detectors/generic.py`, following the same `EnvironmentDetector` pattern as the AWS detector.
- `detect()` returns `True` when `CLOUDSMITH_OIDC_TOKEN` is set and non-empty; `get_token()` returns the value or raises `ValueError` with guidance if unset.
- Registered **last** in `_DETECTORS` so vendor-specific detectors (AWS, etc.) take precedence — the generic detector is a catch-all fallback.
- The token flows through the existing exchange path unchanged.

## Naming note

The environment variable name `CLOUDSMITH_OIDC_TOKEN` was considered against alternatives (`CLOUDSMITH_OIDC_ID_TOKEN`, `CLOUDSMITH_OIDC_JWT`) since "TOKEN" could be misread as a directly-usable API token rather than a JWT to be exchanged. The shorter `CLOUDSMITH_OIDC_TOKEN` was chosen; the `get_token()` error message clarifies it is "the OIDC JWT to exchange for a Cloudsmith token."

## Testing

- Added `cloudsmith_cli/core/tests/test_generic_detector.py` (TDD — written first), mirroring the Bitbucket detector tests: detection (present / unset / empty) and token retrieval (returns / raises).
- All new tests pass; related credential/detector suite (43 tests) passes.

### DIFF
--- a/cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
+++ b/cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING
 from .aws import AWSDetector
 from .azure_devops import AzureDevOpsDetector
 from .base import EnvironmentDetector
+from .generic import GenericDetector
 from .github_actions import GitHubActionsDetector
 
 if TYPE_CHECKING:
@@ -19,6 +20,7 @@ _DETECTORS: list[type[EnvironmentDetector]] = [
     AzureDevOpsDetector,
     GitHubActionsDetector,
     AWSDetector,
+    GenericDetector,
 ]
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Added Azure DevOps to OIDC credential auto-discovery. When running in an Azure DevOps pipeline, the CLI fetches an OIDC token from the `SYSTEM_OIDCREQUESTURI` endpoint using the pipeline's `SYSTEM_ACCESSTOKEN` and exchanges it for a Cloudsmith access token. Works out of the box with no extra dependencies.
 - Added GitHub Actions to OIDC credential auto-discovery. When running in GitHub Actions (with `id-token: write` permission), the CLI fetches an OIDC token from the Actions runtime endpoint and exchanges it for a Cloudsmith access token. Works out of the box with no extra dependencies.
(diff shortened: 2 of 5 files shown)


cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
```diff
@@ -15,6 +16,7 @@
 
 _DETECTORS: list[type[EnvironmentDetector]] = [
     AWSDetector,
+    GenericDetector,
```
BartoszBlizniak, line 23: GenericDetector is registered after AWSDetector, but `AWSDetector.detect()` matches on any ambient boto3 credentials — so a user who explicitly sets `CLOUDSMITH_OIDC_TOKEN` silently gets the AWS STS flow instead. We should perhaps keep generic at the top in case other providers act similarly.

BartoszBlizniak (approved):
pre-apprv

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
cloudsmith-iduffy: This will continue to be a problem.

I'd suggest we add in envvars to individually disable different oidc detectors and another environment variable that enables changes the resolution flow.

cloudsmith-iduffy: Agreed — reordering alone just moves the shadowing problem around (AWS matching ambient boto3 creds is the symptom, not the root). We're deferring detector-selection controls to a dedicated follow-up PR that consolidates #302–#306 and layers two env-var controls on top:

- per-detector disable flags: `CLOUDSMITH_OIDC_<SLUG>_DISABLED` (e.g. `CLOUDSMITH_OIDC_AWS_DISABLED=true`)
- an ordered evaluation list: `CLOUDSMITH_OIDC_DETECTOR_ORDER=generic,aws,...`

Precedence: the order list defines the candidate set + sequence; the `*_DISABLED` flags filter on top, so disable always wins. Keeping this PR scoped to just the generic detector so it can merge; the controls land once all detectors are in.

cloudsmith-iduffy: I'm going to follow this up in a separate PR once everything is merged in.

PR #309, opened 2026-06-10
fix(security): resolve code scanning alerts in JWT and checksum handling

BartoszBlizniak:
# Description

Resolves code scanning alerts:

- `oidc/cache.py`: keeps PyJWT's `jwt.decode(..., options={"verify_signature": False})` per review, documents why the signature is deliberately not verified (the `exp` claim is only used for a cache TTL on our own server-issued token, never for authorization), and drops the `algorithms` list — it has no effect when verification is disabled and was flagged separately by code scanning.
- `utils.py` / `download.py`: marks MD5/SHA1 checksum hashing with `usedforsecurity=False` — used solely to verify API-provided package checksums, not for cryptography.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

The remaining "Unverified JWT Token Decoding" alert is dismissed as won't fix with justification: the decode is intentional and never feeds an authorization decision. Full pytest green on Python 3.10 and 3.14. Live push/download round-trip exercised the checksum paths. `usedforsecurity=False` requires Python ≥3.9, below the project's 3.10 floor.


cloudsmith_cli/core/credentials/oidc/cache.py
```diff
@@ -41,15 +42,16 @@ def _cache_key(api_host: str, org: str, service_slug: str) -> str:
...
 
-        payload = jwt.decode(
-            token,
-            options={"verify_signature": False},
-            algorithms=["RS256", "ES256", "HS256"],
-        )
+    The token is only inspected to determine a cache TTL; it is never used to
+    make an authorization decision, so the signature is deliberately not
+    verified (the API rejects tampered tokens regardless).
+    """
```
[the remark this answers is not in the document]
BartoszBlizniak: I'll make the changes for this comment in #308 to keep dependency changes in one PR

cloudsmith_cli/core/credentials/oidc/cache.py
```diff
@@ -41,15 +42,19 @@ def _cache_key(api_host: str, org: str, service_slug: str) -> str:
...
+        return None
     try:
-        import jwt
-
-        payload = jwt.decode(
-            token,
-            options={"verify_signature": False},
-            algorithms=["RS256", "ES256", "HS256"],
-        )
+        payload_segment = parts[1]
```
cloudsmith-iduffy: Don't agree with AIs take on this one, prefer the old approach rather than having to understand the make up of a JWT ourselves.

cloudsmith-iduffy (approved):
Approving for the usedforsecurity=False but the JWT thing should be reverted, best to let a library handle understanding how a JWT is decoded.

cloudsmith_cli/core/credentials/oidc/cache.py
BartoszBlizniak: Looks like it was trying to solve: https://github.com/cloudsmith-io/cloudsmith-cli/security/code-scanning/38

PR #311, opened 2026-06-10
feat: add OIDC detector enable/order controls

cloudsmith-iduffy:
## What

Adds **controls for OIDC detector selection** to the credential auto-discovery chain, so a detector that matches an environment you don't want it to (for example, the AWS detector matching ambient instance credentials) can be disabled or reordered.

## Controls

Both controls are resolved through the credential context — the same plumbing as the other `--oidc-*` options — so `detect_environment()` reads only from the context, never directly from the environment.

- **Disable a detector** — `CLOUDSMITH_OIDC_<ID>_DISABLED=true` skips it. Only the literal string `true` (case-insensitive) disables; anything else leaves it enabled. The credentials decorator resolves these into `context.oidc_disabled_detectors`.
- **Reorder evaluation** — `--oidc-detector-order` (or the `CLOUDSMITH_OIDC_DETECTOR_ORDER` env var) takes a comma-separated list of detector ids and overrides both which detectors are considered and the order they are tried in (first match wins). Ids not listed are skipped; unknown ids are logged at debug and ignored.

**Precedence:** the order list defines the candidate set and sequence, then the disabled set is applied on top — a disabled detector is always skipped even if it appears in the order list. With neither set, behaviour is identical to today's fixed priority order.

Detector ids: `aws`, `azure_devops`, `bitbucket`, `circleci`, `generic`, `github`, `gitlab`.

## How

- Each detector class gains a stable `id` attribute; the `EnvironmentDetector` base gets `id = "base"`.
- `detect_environment()` selects via `_ordered_detectors(order)` + `_enabled_detectors(order, disabled)`; a public `registered_detectors()` accessor is added; `disabled_detectors_from_env()` resolves the per-detector disable env vars.
- New `CredentialContext` fields `oidc_detector_order` and `oidc_disabled_detectors`, wired through `decorators.py` (the `--oidc-detector-order` option) and `config.py`, following the existing OIDC-option chain.
- README + CHANGELOG document both controls and the precedence rule.

## Testing

- New `cloudsmith_cli/core/tests/test_detector_controls.py` (TDD): default order, per-detector disable (via context), env-var truthiness (`true` only), env resolution + var naming, order reordering, order limiting the candidate set, unknown-id-ignored, empty-order fallback, order+disable composition, and a registry id-uniqueness guard.
- Full relevant suite green; `black`/`flake8`/`isort`/`pyupgrade`/`pylint` (10/10)/`typos` all pass.

## Note for reviewers

Google Cloud (`gcp`) is **not** part of this PR — it isn't in `master` yet. #306 layers the GCP detector (and its `id = "gcp"`) on top of this branch and is targeted at it.


cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
```diff
@@ -3,32 +3,99 @@
...
+        detector_cls = detectors_by_slug.get(slug)
+        if detector_cls is None:
+            logger.debug(
+                "Ignoring unknown OIDC detector slug in %s: %s",
+                DETECTOR_ORDER_ENV_VAR,
+                slug,
+            )
+            continue
+        ordered.append(detector_cls)
+    return ordered
```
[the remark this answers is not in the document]
cloudsmith-iduffy: Fixed in 61dc52d — duplicate ids in the order list now keep their first position and each detector is evaluated at most once (covered by `test_duplicate_ids_run_each_detector_once`). Also added a debug log when the order/disable controls leave no detectors enabled.

cloudsmith_cli/core/tests/test_gcp_detector.py
```diff
@@ -0,0 +1,190 @@
...
+            AzureDevOpsDetector,
+            GitHubActionsDetector,
+        )
+
+        assert _DETECTORS == [
+            AzureDevOpsDetector,
+            GitHubActionsDetector,
+            AWSDetector,
+            GCPDetector,
+        ]
```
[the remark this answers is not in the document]
cloudsmith-iduffy: This comment is anchored to an earlier version of the branch — `test_gcp_detector.py` is no longer part of this PR. The GCP detector (and its tests) was split out to #306, which layers on top of this branch; the relative-ordering suggestion applies there.

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
```diff
@@ -30,11 +32,77 @@
...
+
+    detectors_by_id = {detector_cls.id: detector_cls for detector_cls in _DETECTORS}
+    ordered: dict[str, type[EnvironmentDetector]] = {}
+    for token in raw_order.split(","):
+        identifier = token.strip().lower()
+        if not identifier or identifier in ordered:
+            continue
+        detector_cls = detectors_by_id.get(identifier)
+        if detector_cls is None:
+            logger.debug("Ignoring unknown OIDC detector id: %s", identifier)
```
BartoszBlizniak, line 79: Should we bump this to `click.UsageError` and surface as a warning instead?

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
```diff
@@ -30,11 +32,77 @@
...
 def detect_environment(
     context: CredentialContext,
 ) -> EnvironmentDetector | None:
     """Try each detector in order, returning the first that matches."""
-    for detector_cls in _DETECTORS:
+    enabled = _enabled_detectors(
+        context.oidc_detector_order, context.oidc_disabled_detectors
+    )
+    if not enabled:
+        logger.debug("No OIDC detectors enabled after applying order/disable controls")
```
BartoszBlizniak, line 104: Same as above here

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
BartoszBlizniak, line 31: Should we move the Generic detector to be first in order of detection here or would it not matter too much?

README.md
```diff
@@ -211,6 +211,15 @@ See the [Cloudsmith GitLab CI/CD integration guide](https://docs.cloudsmith.com/
...
 
+#### Controlling OIDC Detector Selection
+
+By default the CLI tries each detector in a fixed priority order and uses the first that matches. Two controls let you override this when a detector matches an environment you don't want it to (for example, the AWS detector matching ambient instance credentials):
+
+- **Disable a detector** — set `CLOUDSMITH_OIDC_<DETECTOR>_DISABLED=true` to skip it entirely. Only the literal value `true` (case-insensitive) disables; anything else leaves the detector enabled. For example, `CLOUDSMITH_OIDC_AWS_DISABLED=true` skips the AWS detector so an explicitly-set `CLOUDSMITH_OIDC_TOKEN` is picked up by the generic detector instead.
```
BartoszBlizniak, line 218: Should we also allow this in the config file? 

i.e.:

```
  [default]
  oidc_disabled_detectors = aws,gcp
  oidc_detector_order = github, azure_devops
```

BartoszBlizniak (approved):
Few comments there but overall happy to merge in 👍

README.md
cloudsmith-iduffy: Done in 7906cdc — both controls are now settable in `config.ini` (under `[default]` or a profile):

```ini
[default]
oidc_detector_order = github, aws
oidc_disabled_detectors = aws, gitlab
```

`oidc_disabled_detectors` is additive with the per-detector `CLOUDSMITH_OIDC_<ID>_DISABLED` env vars; `--oidc-detector-order` / `CLOUDSMITH_OIDC_DETECTOR_ORDER` still override the config order. README + CHANGELOG updated, covered by `test_oidc_detector_config.py`.

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
cloudsmith-iduffy: Surfaced this as a **warning** rather than `click.UsageError`, and moved it to the CLI layer (`decorators._warn_on_oidc_detector_controls`) instead of here in core (7906cdc). Reasoning:

- The OIDC provider is one best-effort link in the credential fallback chain — it logs and returns `None` so resolution can fall through to an API key. Raising `UsageError` from `detect_environment` would hard-abort the whole command on a typo, even when another credential source would have worked.
- An unknown id is not always a mistake: `gcp` is a valid id once #306 merges but not before, so a shared config/env listing it should not fail on an older CLI.
- The detectors module is deliberately click-free; keeping the warning in the CLI layer preserves that.

The warning now also lists the valid ids to help fix typos.

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
cloudsmith-iduffy: Same treatment — now a non-fatal CLI-layer warning ("no detectors are enabled after applying the detector order/disable controls; OIDC auto-discovery will be skipped") in `_warn_on_oidc_detector_controls`, so the credential chain still falls back rather than aborting. 7906cdc

cloudsmith_cli/core/credentials/oidc/detectors/__init__.py
cloudsmith-iduffy: Kept generic last for now. It only matches when `CLOUDSMITH_OIDC_TOKEN` is explicitly set, so on platforms that do not use that var, moving it first changes nothing — it would only change behaviour when the var is set, and that case is already served by the new reorder/disable controls (the README documents using `CLOUDSMITH_OIDC_AWS_DISABLED` / order to make the explicit token win). Leaving the platform-specific detectors first keeps generic a true fallback and gives more accurate source labels (e.g. "GitLab CI" vs "Generic"). Happy to flip it to explicit-token-wins as the default if you would prefer that.

PR #317, opened 2026-06-17
build(ENG-12825): migrate packaging to uv + pyproject.toml

BartoszBlizniak:
# Description

Migrate CLI packaging to `pyproject.toml` + `uv` (with a locked `uv.lock`) as a single source of truth, replacing `setup.py`, `setup.cfg`, and `requirements.{in,txt}`. CI (`test.yml`, `lint.yml`) now runs through `uv`. No change to CLI behaviour.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring
- [ ] Other (please describe)

## Additional Notes

- Build/packaging only — the CLI is unchanged; `uv build` produces the same wheel + sdist.
- `requirements.txt` is removed; `uv export` regenerates a pinned list if a consumer needs one.
- Groundwork for distributing the CLI as a standalone binary.


pyproject.toml
```diff
@@ -0,0 +1,138 @@
...
+requires = ["setuptools==82.0.1", "wheel==0.47.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "cloudsmith-cli"
+dynamic = ["version"]
+description = "Cloudsmith Command-Line Interface (CLI)"
+readme = "README.md"
+license = "Apache-2.0"
+license-files = ["LICENSE"]
```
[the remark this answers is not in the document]
BartoszBlizniak: Verified empirically — keeping the SPDX string. `uv build` succeeds with `license = "Apache-2.0"` and emits `License-Expression: Apache-2.0` in the wheel METADATA. PEP 639 / setuptools ≥77 supports the SPDX expression string; the `{ file = ... }` / `{ text = ... }` table forms are the deprecated style.

pyproject.toml
```diff
@@ -0,0 +1,138 @@
...
+    # 'Development Status :: 3 - Alpha',
+    # 'Development Status :: 4 - Beta',
+    "Development Status :: 5 - Production/Stable",
+    # 'Development Status :: 6 - Mature',
+    # 'Development Status :: 7 - Inactive',
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "Intended Audience :: System Administrators",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python",
```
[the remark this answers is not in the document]
BartoszBlizniak: Verified — not re-adding this. With setuptools (PEP 639), re-adding `License :: OSI Approved :: Apache Software License` makes `uv build` fail: `InvalidConfigError: License classifiers have been superseded by license expressions`. The license is already conveyed via `License-Expression: Apache-2.0` in the wheel METADATA.

.github/workflows/release.yml
```diff
@@ -33,6 +33,11 @@ jobs:
                   python -m pip install --upgrade pip
                   pip install pex setuptools wheel
 
+            - name: Set up uv
+              uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+              with:
+                  enable-cache: false
```
cloudsmith-iduffy, line 40: Why no cache?

cloudsmith_cli/cli/utils.py
```diff
@@ -194,6 +194,25 @@ def maybe_truncate_list(data, max_len=5):
     return data
 
 
+def maybe_unstyle_prompt(prompt, err=False):
```
cloudsmith-iduffy, line 197: This seems unrelated to the moving to uv and pyproject.toml stuff. Looking at the commits it seems like we were forced to do it due to the click upgrade.

cloudsmith-iduffy (approved):
Two small questions but nothing blocking.

.github/workflows/release.yml
BartoszBlizniak: Zizmor's cache-poisoning audit flags restoring caches in tag-triggered publish workflows. uv is only used for `uv build` here so the cache saves nothing anyway. Added a comment in the file to note this.

cloudsmith_cli/cli/utils.py
BartoszBlizniak: Correct — the uv re-lock moved click to 8.4.1 and click 8.4.0 changed prompts to go straight through input() (for readline) instead of echo(). Without the workaround the tests asserting the plain confirm-prompt text fail.

PR #333, opened 2026-07-30
feat(ENG-13681): add credential-helper domains command

cloudsmith-iduffy:
## Description

Adds `cloudsmith credential-helper domains`, which answers the question `credential-helper generic` (#332) deliberately does not: **which hosts can Cloudsmith authenticate?**

```console
$ CLOUDSMITH_ORG=acme cloudsmith credential-helper domains
{"version": 1, "domains": [{"host": "python.cloudsmith.io", "format": "python",
  "backend_kind": 3, "enabled": true, "validated": true, "type": "default"}, ...]}
```

A consumer needs both commands: this one to decide whether a host should receive a credential, `generic` to get it. Splitting them this way means a keyring backend shells out for a token only on a host it has already confirmed.

### Why the built-in hosts are included

Custom domains are not knowable without an API call, but the standard service hosts are. Shipping the built-in table **here** rather than letting each consumer hardcode `*.cloudsmith.io` means there is exactly one place that knows the host list, so consumers cannot drift out of sync with the platform. 21 built-in hosts are listed, and listing them needs no organisation and no authentication.

The command takes no arguments. The organisation for the custom-domain lookup resolves from click context — `CLOUDSMITH_ORG` or `oidc_org` in `config.ini`, which anyone authenticating via SAML has typically already set — and with no organisation configured the built-ins are still listed. Consumers that want only custom domains filter on `type`.

Disabled and unvalidated custom domains are listed too, with their state intact — a domain that exists but does not work should be visible rather than absent, since "why doesn't my domain work?" is the main reason to run this.

### Machine-only

Like `generic`, this is for programmatic consumers and has no human-readable output mode: always one JSON document on stdout, no table, no `-F`. Warnings go to stderr so stdout stays parseable.

### Config override, and why it is restricted

The built-in table can be replaced wholesale by a `[domains]` section in `config.ini`, for dedicated and on-premise deployments.

That override is honoured **only from trusted configuration** — your user-level config directory, `~/.cloudsmith`, or an explicit `--config-file`. `config.ini` is searched in the working directory first (`cli/config.py:48-52`), so a repository can ship one, and this command's output decides which hosts may receive a credential. Honouring a directory-relative section would let a checked-in `config.ini` declare `evil.example.com` a Cloudsmith host and harvest a live token — the same class of vector `_guard_untrusted_endpoints` (`cli/decorators.py:262`) already closes for `api_host`. A directory-relative section is ignored with a warning on stderr rather than silently.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

Stacked on #332. Full suite passes (661 passed, 40 skipped), with 41 tests across the command and the domain table — including the security property that a working-directory `config.ini` can never influence the host list.

The shellplugin's hardcoded `DEFAULT_CDN_HOST`/`DEFAULT_UPLOAD_HOST` now derive from the shared table via the new `CDN_HOST` constant and `builtin_host()` lookup, so the table genuinely is the single place that knows the service hosts.

`dl` and `upload` carry no `backend_kind` because they serve every format; that matches the server, which rejects a `backend_kind` on a download-type domain.


BartoszBlizniak (approved):
nice one @cloudsmith-iduffy 🦾

PR #337, opened 2026-08-01
feat(ENG-13681): add `cloudsmith domains list` with custom-domain discovery

cloudsmith-iduffy:
# Description

Adds `cloudsmith domains list`, which emits a versioned JSON document listing
the hosts Cloudsmith can authenticate — built-in `*.cloudsmith.io` service hosts
plus an organisation's custom domains.

- **Filters on the domain list.** `--format` keeps the hosts usable for a package format: a host
  serving no single format serves every one, so the download CDN and the
  generic upload endpoint survive any `--format`. `--repo` keeps the hosts
  usable for a repository, most-preferred first. `--domain-type` is an exact match instead — a host has
  exactly one purpose — so pairing it with `--format` asks for a host
  speaking a format's own protocol.

- **`--org` is the name of the organisation option.** `--organization` and
  `--oidc-org` remain as aliases, and `org`, `organization` or `oidc_org` are
  all accepted in `config.ini`. `--oidc-org` named the setting after the first
  feature that wanted it; custom-domain discovery reads it too, so it is now
  named after what it is. `CLOUDSMITH_ORG` is unchanged, and
  `credential-helper install` no longer carries a separate `--org` of its own.

- **A replaceable domain table.** A dedicated deployment, serving packages from
  its own domains, can replace the built-in hosts via a `[domains]` section in
  `config.ini`. Each entry names the package format its host serves, or
  `download`/`upload` for the two endpoints no format names; an entry naming
  nothing recognisable is skipped with a warning rather than guessed at. This is
  honoured only from trusted configuration — your config directory,
  `~/.cloudsmith`, or an explicit `--config-file`. A `[domains]` section in a
  directory-relative `config.ini` is ignored with a warning, because that file
  can be committed to a repository and this list decides which hosts may receive
  a credential.

- **Server-matching precedence between overlapping domains.** A domain bound to
  the repository in hand beats an organisation-wide one, `primary` beats
  secondary, and the oldest breaks any remaining tie. With `--repo` the custom
  domains come back most-preferred first, so the first entry is the one
  Cloudsmith would bind for that repository.

- **A seven-day on-disk cache**, with `--refresh` to bypass it

- With no organisation configured, the command lists whatever custom domains an
  earlier run already cached, at no API cost, rather than enumerating every
  organisation the caller belongs to.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

### DIFF
--- a/cloudsmith_cli/cli/commands/domains.py
+++ b/cloudsmith_cli/cli/commands/domains.py
@@ -12,8 +12,10 @@ import json
 import os
 
 import click
+from click.core import ParameterSource
 
 from ...core.api.exceptions import ApiException
+from ...core.pagination import PageInfo
 from ...credential_helpers.custom_domains import (
     get_custom_domains,
     order_by_precedence,
@@ -30,6 +32,7 @@ from ..config import get_default_config_path
 from ..decorators import (
     common_api_auth_options,
     common_cli_config_options,
+    common_cli_list_options,
     initialise_api,
 )
 from .main import main
@@ -109,6 +112,7 @@ def domains_():
     "Drops custom domains bound to a different repository.",
 )
 @common_cli_config_options
+@common_cli_list_options
 @common_api_auth_options
 @initialise_api
 @click.pass_context
@@ -119,6 +123,9 @@ def list_domains(  # pylint: disable=too-many-arguments
     format_: str | None,
     domain_type: str | None,
     repo: str | None,
+    page: int,
+    page_size: int,
+    page_all: bool,
 ) -> None:
     """Emit the Cloudsmith hosts and their package formats as JSON.
 
@@ -147,10 +154,21 @@ def list_domains(  # pylint: disable=too-many-arguments
     Cloudsmith would bind; ``--repo`` is what lets the repository-bound hosts
     take part in it.
 
+    Results support the same --page/--page-size (-p/-l) options as other
+    `list` commands, selecting a page of the combined, filtered list. Unlike
+    those commands, this one defaults to --page-all (--show-all): the
+    combined list is small and built from a local cache rather than a live
+    paginated endpoint, so there is no benefit to hiding entries by default.
+    Passing --page or --page-size switches to paged output.
+
     Output (stdout):
         JSON: {"version": 1, "domains": [{"host": ..., "format": ...,
         "type": ..., "domain_type": ..., "org": ..., "repository": ...,
-        "primary": ..., "created_at": ...}]}
+        "primary": ..., "created_at": ...}], "meta": {"pagination": {...}}}
+
+        "meta" is only present when the result is paginated (i.e. not
+        --page-all), and mirrors the pagination metadata other `list`
+        commands emit for `-F json`.
 
     Examples:
 
@@ -170,6 +188,24 @@ def list_domains(  # pylint: disable=too-many-arguments
         # Where to upload to
         $ cloudsmith domains list --org my-org --domain-type upload
     """
+    if not page_all:
+        explicit_sources = {
+            src
+            for src in (
+                ParameterSource.COMMANDLINE,
+                ParameterSource.ENVIRONMENT,
+                getattr(ParameterSource, "PROMPT", None),
+            )
+            if src is not None
+        }
+        page_explicit = ctx.get_parameter_source("page") in explicit_sources
+        page_size_explicit = ctx.get_parameter_source("page_size") in explicit_sources
+        if not page_explicit and not page_size_explicit:
+            # Unlike other `list` commands, domains list defaults to showing
+            # everything: the combined list is small and built from a local
+            # cache, so paging by default would hide entries for no benefit.
+            page_all = True
+
     explicit_config = ctx.meta.get("config_file")
     if explicit_config and os.path.isdir(explicit_config):
         explicit_config = os.path.join(explicit_config, "config.ini")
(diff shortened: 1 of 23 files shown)


cloudsmith_cli/cli/commands/domains.py
```diff
@@ -0,0 +1,244 @@
...
+    domain_type: str | None,
+    repo: str | None,
+) -> None:
+    """Emit the Cloudsmith hosts and their package formats as JSON.
+
+    Machine-readable output for programmatic consumers. It lists the built-in
+    ``*.cloudsmith.io`` service hosts alongside custom domains, with ``type``
+    distinguishing ``default`` from ``custom`` and ``domain_type`` saying what
+    each host is for: ``download``, ``upload``, or ``native_api`` for a host
+    speaking one package format's own protocol. An organisation's own custom
```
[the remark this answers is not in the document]
cloudsmith-iduffy: Fixed — the docstring now lists all four purposes.

Worth adding that `api` is genuinely reachable, so the docstring was the wrong half: `SERVER_DOMAIN_TYPES` maps server value `2` to `DomainType.API`, so a server-side custom domain can carry it. No built-in host does, which is probably why it was missed — and that turned out to be its own defect (`--domain-type api` matched nothing at all). The option help and the docstring now both say explicitly that no built-in host is an `api` host and only a custom domain can be.

cloudsmith_cli/credential_helpers/docker/installer.py
```diff
@@ -131,11 +131,7 @@ def install(
 
         # --- Custom-domain auto-discovery (best-effort) ---
         if discover:
-            if dry_run:
-                # Discovery queries the API and refreshes the on-disk domain
-                # cache, neither of which a "no changes" preview may do.
-                actions.append("skipped custom-domain auto-discovery (dry run)")
-            elif org and credential and credential.api_key:
+            if org and credential and credential.api_key:
```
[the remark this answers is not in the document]
cloudsmith-iduffy: You were right, and it is a regression rather than a design decision — thanks for catching it.

I initially read this as intentional, on the grounds that the preview enumerates `would set credHelpers['<host>']=...` for every host and so needs the discovered ones to be accurate. `git show 054a88c` says otherwise: the `if dry_run:` short-circuit was deleted in this commit, along with the test that pinned it ("a dry run must not query the custom-domain API"). The original even carried the rationale you gave — "Discovery queries the API and refreshes the on-disk domain cache, neither of which a 'no changes' preview may do."

Restored the guard and both deleted test scenarios (`dry_run` and `blank_credential`, which went in the same edit), and reverted the docstring change I had made to rationalise the new behaviour.

cloudsmith_cli/credential_helpers/custom_domains.py
```diff
@@ -253,13 +346,19 @@ def get_format_domains(
         refresh: When ``True``, bypass the cache and fetch fresh data from the API.
 
     Returns:
-        List of hostnames that are enabled, validated, and match the given backend_kind.
+        Hostnames that are usable and match the given backend_kind, in
+        precedence order - the host Cloudsmith would treat as active first.
```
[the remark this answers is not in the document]
cloudsmith-iduffy: Fixed — `configure_api` is now threaded through `get_format_domains`, with a test.

Confirming the mechanism you describe is real: `initialise_api` constructs a fresh `Configuration()` and `set_default()`s it, so a narrower re-init does discard proxy, `verify_ssl`, user-agent, headers and the retry settings.

One correction to the premise, though: no current caller is affected. Both `get_format_domains` callers sit under `credential-helper docker` and `credential-helper install`, and neither command uses `@initialise_api` — they use `@common_api_auth_options` + `@resolve_credentials` — so `configure_api=True` is precisely what configures the SDK on those paths. The only place that needs `False` today is `domains list`, which already calls `get_custom_domains` directly. So this closes a latent footgun for a future caller rather than fixing a live bug.

cloudsmith_cli/credential_helpers/default_domains.py
```diff
@@ -0,0 +1,305 @@
...
+    backend_kind: int | None
+    domain_type: DomainType = DomainType.NATIVE_API
+
+    @property
+    def format_label(self) -> str | None:
+        """The package format this host serves, or None for no single format."""
+        return format_for_backend_kind(self.backend_kind)
+
+
+BUILTIN_DOMAINS: tuple[DefaultDomain, ...] = (
```
BartoszBlizniak, line 105: Unrelated here, but it would be awesome if we had an endpoint for listing all default domains

cloudsmith_cli/cli/commands/domains.py
```diff
@@ -0,0 +1,246 @@
...
+    "repo",
+    default=None,
+    help="Only list hosts usable for this repository, most-preferred first. "
+    "Drops custom domains bound to a different repository.",
+)
+@common_cli_config_options
+@common_api_auth_options
+@initialise_api
+@click.pass_context
+def list_domains(  # pylint: disable=too-many-arguments
```
BartoszBlizniak, line 119: nit; Do we want to add pagination here (if the endpoint supports it)? Unlikely a customer will have that many domains, but it keeps consistent with the cli flags.

BartoszBlizniak (approved):
one nit but lgtm

cloudsmith_cli/credential_helpers/default_domains.py
cloudsmith-iduffy: Yes! that would make a lot of sense, I will create a ticket for that and its likely something our group can take on.

cloudsmith_cli/cli/commands/domains.py
cloudsmith-iduffy: Will look into it.

cloudsmith-iduffy: added.

PR #338, opened 2026-08-01
feat(ENG-13681): add credential-helper generic command

cloudsmith-iduffy:
# Description

Stack 3/4. Base: #337.

Adds `cloudsmith credential-helper generic`, which resolves a credential through
the full provider chain (API key, `credentials.ini`, system keyring, OIDC) and
emits it as a versioned JSON document —
`{"version": 1, "username": "token", "password": "<token>"}` — for tools that
shell out to the CLI rather than importing it.

It takes no arguments: a Cloudsmith token is organisation-wide, so the host it
will be used against does not change which credential resolves. Errors exit
non-zero with a message on stderr and never emit a partial document.

Independent of the custom-domain work below it in the stack; it sits here only
to keep the stack linear.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

Test suite: 659 passed, 40 skipped at this commit.


BartoszBlizniak:
@cloudsmith-iduffy - I'll switch this from draft to ready for review and take a look, does what #341 needed to do.

BartoszBlizniak (approved):

PR #339, opened 2026-08-01
feat(ENG-13683): add Maven credential helper

cloudsmith-iduffy:
# Description

The last of the stack, rebased onto master now that #336, #337 and #338 have
landed, and squashed to a single commit.

Maven has no credential-helper protocol, so this authenticates it with a shell
plugin instead. `credential-helper install maven --org <org> --repo <repo>`
writes an `mvn` shim that wraps every invocation in `cloudsmith exec`, which
injects an ephemeral mode-0600 `settings.xml` via `mvn -s` and deletes it when
the run ends — no token is ever written to durable configuration.

- `credential-helper shell-init` prints the shell initialisation (bash, zsh,
  fish) that puts the shims directory first on `PATH`.
- `cloudsmith exec -- <command>` is callable directly, for CI that should not
  touch `PATH`.
- Publishing is opt-in: `install` prints the `distributionManagement` snippet to
  add to `pom.xml`.
- Custom download/upload domains are discovered from the organisation at
  install time. Domains bound to a single repository are left out — they need
  URLs of a different shape, which is its own change — so a repository-scoped
  domain falls back to the default host. Which kind each host is gets recorded
  in `package-managers.ini`, so wrapped runs need no lookup — the run path
  makes no API call and no cache read.

**Credential binding.** One `<server>` id covers download and upload. It
defaults to the literal `cloudsmith`, matching the `distributionManagement`
snippet `install` prints, so a team can share one `pom.xml`. Maven matches a
server's credentials to a repository by id alone, with no host check, so a
checked-out `pom.xml` declaring a repository under that id receives the token
on an ordinary `mvn compile` — the same exposure as the `~/.m2/settings.xml` a
Maven user would otherwise keep. Pass `--server-id` at install time to bind
the credential to an id a third-party `pom.xml` cannot guess; minting a random
id by default would break a team sharing one `pom.xml`, so it is opt-in.
Documented in the CHANGELOG's Security section and `settings.py`'s module
docstring.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

Test suite: 682 passed, 40 skipped at this commit.

An earlier, larger draft of this branch was verified end to end in GitHub
Actions with OIDC as the only credential
([run](https://github.com/cloudsmith-iduffy/cloudsmith-cli-oidc-test/actions/runs/31288628736)):
a plain `mvn -B clean deploy` resolved a private dependency through the
organisation's discovered custom domain and published the built jar to
`maven.cloudsmith.io/iduffy-demo/default/`. That draft carried a cached
`settings.xml`, a shell fast path and deploy-goal detection; this PR ships
without them to halve the diff, and the shim path itself is unchanged.

Two things a reviewer should know:

- Wrapped runs do not consult `~/.m2/settings.xml`; mirrors, proxies and other
  `<server>` entries are not seen. Passing your own `-s/--settings` runs Maven
  unwrapped, with a warning. This is the designed mechanism — Maven has no
  settings-merge — and is called out at install time and in the CHANGELOG.
- One known limitation ships knowingly. The ephemeral `settings.xml` is removed
  in a `finally`, so a SIGKILL/SIGTERM of `cloudsmith exec` (a cancelled CI step)
  leaves the 0600 token file in `/tmp` until the agent is recycled; a signal
  handler is the fix and wants its own change.

### DIFF
--- a/cloudsmith_cli/cli/commands/credential_helper/shell.py
+++ b/cloudsmith_cli/cli/commands/credential_helper/shell.py
@@ -1,0 +1,55 @@
+# Copyright 2026 Cloudsmith Ltd
+"""``cloudsmith credential-helper shell-init`` — print shell init for shims.
+
+Add ``eval "$(cloudsmith credential-helper shell-init)"`` to your shell rc file
+to put the Cloudsmith shims directory ahead of the real package-manager
+binaries on ``$PATH``.
+"""
+
+import os
+
+import click
+
+from ....credential_helpers.maven.config import shims_dir
+
+_COMMENT = "# Put Cloudsmith package-manager shims ahead of the real binaries"
+
+_POSIX_STATEMENT = 'export PATH="{path}:$PATH"'
+
+_STATEMENTS = {
+    "bash": _POSIX_STATEMENT,
+    "zsh": _POSIX_STATEMENT,
+    "fish": 'fish_add_path "{path}"',
+}
+
+
+def detect_shell():
+    """Best-effort shell detection from ``$SHELL``, defaulting to bash."""
+    name = os.path.basename(os.environ.get("SHELL", ""))
+    return name if name in _STATEMENTS else "bash"
+
+
+@click.command(name="shell-init")
+@click.option(
+    "--shell",
+    "shell_name",
+    type=click.Choice(sorted(_STATEMENTS)),
+    default=None,
+    help="Target shell. Auto-detected from $SHELL when omitted.",
+)
+def shell_init(shell_name):
+    """Print shell init that puts the Cloudsmith shims dir first on PATH.
+
+    Examples:
+
+    \b
+        # bash / zsh
+        $ eval "$(cloudsmith credential-helper shell-init)"
+
+    \b
+        # fish
+        $ cloudsmith credential-helper shell-init --shell fish | source
+    """
+    statement = _STATEMENTS[shell_name or detect_shell()]
+    click.echo(_COMMENT)
+    click.echo(statement.format(path=shims_dir()))
--- a/cloudsmith_cli/cli/commands/exec_.py
+++ b/cloudsmith_cli/cli/commands/exec_.py
@@ -1,0 +1,34 @@
+# Copyright 2026 Cloudsmith Ltd
+"""CLI/Commands - Run a command with Cloudsmith credentials provisioned."""
+
+import sys
+
+import click
+
+from ...credential_helpers.maven import runner
+from ..decorators import common_api_auth_options, resolve_credentials
+from .main import main
+
+
+@main.command(name="exec", context_settings={"ignore_unknown_options": True})
+@click.argument("command", nargs=-1, type=click.UNPROCESSED, required=True)
+@common_api_auth_options
+@resolve_credentials
+@click.pass_context
+def exec_(ctx, opts, command):
+    """Run a package-manager command authenticated against Cloudsmith.
+
+    Wraps the command so it resolves dependencies from your Cloudsmith
+    repository, with credentials injected for that run and removed afterwards.
+    This is the machinery the ``mvn`` shim uses, callable directly in CI
+    without touching ``PATH``.
+
+    Maven runs use a generated ``settings.xml``; your ``~/.m2/settings.xml``
+    is not consulted. The repository comes from ``credential-helper install
+    maven``. The package manager is detected from the command, so just put it
+    after ``--``:
+
+    \b
+        $ cloudsmith exec -- mvn clean install
+    """
+    sys.exit(runner.run(list(command), credential=opts.credential))
--- a/cloudsmith_cli/credential_helpers/maven/installer.py
+++ b/cloudsmith_cli/credential_helpers/maven/installer.py
@@ -1,0 +1,262 @@
+# Copyright 2026 Cloudsmith Ltd
+"""Installer for the Maven credential helper.
+
+Writes an ``mvn`` shim that re-execs ``cloudsmith exec -- mvn "$@"`` into the
+Cloudsmith shims dir, records the repository binding, and prints the
+``distributionManagement`` snippet needed for ``mvn deploy``.  Dependency
+resolution works transparently once the shims dir is on PATH (via
+``credential-helper shell-init``); publishing is opt-in.
+
+Maven uses two distinct endpoints, so two custom-domain kinds matter:
+- **download** (dependency resolution) goes via the download CDN; its custom
+  domains carry ``backend_kind is None`` (the generic download domain).
+- **upload** (``distributionManagement``) goes via the native Maven endpoint;
+  its custom domains carry ``BackendKind.MAVEN``.
+"""
+
+from __future__ import annotations
+
+import logging
+import shutil
+from xml.sax.saxutils import escape
+
+from ...core.api.exceptions import ApiException
+from ...core.credentials.models import CredentialResult
+from ...templates import render
+from ..backends import BackendKind
+from ..custom_domains import CustomDomain, get_custom_domains, order_by_precedence
+from ..default_domains import DomainType
+from ..launchers import (
+    cloudsmith_command,
+    is_frozen,
+    is_on_path,
+    launcher_filename,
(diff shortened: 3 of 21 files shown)


cloudsmith_cli/credential_helpers/maven/runner.py
```diff
@@ -0,0 +1,181 @@
...
+import sys
+import tempfile
+
+from . import config, settings
+
+BINARY_NAMES = ("mvn", "mvnw")
+
+# Flags that short-circuit Maven wherever they appear, so there is nothing to
+# authenticate.
+_SKIP_AUTH_ARGS = frozenset({"--help", "-h", "--version", "-v", "help"})
```
BartoszBlizniak, line 26: Bare `"help"` in [`_SKIP_AUTH_ARGS`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/4a1cb9d4ad628e09f8ab118af4fedd812a02f16d/cloudsmith_cli/credential_helpers/maven/runner.py#L26) matches anywhere in the args, so option *values* trip it: `mvn -pl help clean install` (a module named `help`) or `mvn -P help ...` silently runs without credential injection and fails on the first private dependency, with nothing pointing at the shim.

Since bare `mvn help` isn't a valid Maven 3 invocation anyway (`help` is only usable as a `help:<goal>` prefix, which this doesn't match), the entry buys nothing - I'd drop it and keep just the `-h/--help/-v/--version` flags.

cloudsmith_cli/credential_helpers/maven/runner.py
```diff
@@ -0,0 +1,181 @@
...
+    """Run Maven against a generated ``settings.xml``, deleted afterwards."""
+    token = credential.api_key if credential else None
+    if not token:
+        print(
+            "cloudsmith: warning: no credential resolved; private repositories "
+            "will fail to authenticate — set CLOUDSMITH_API_KEY or configure OIDC",
+            file=sys.stderr,
+        )
+
+    temp_dir = tempfile.mkdtemp(prefix="cloudsmith-maven-")
```
BartoszBlizniak, line 167: On the acknowledged settings.xml-outlives-a-killed-run limitation ([`_run_with_settings`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/4a1cb9d4ad628e09f8ab118af4fedd812a02f16d/cloudsmith_cli/credential_helpers/maven/runner.py#L167)): I reproduced it - `kill -9` of `cloudsmith exec` mid-run leaves the 0600 token file in `/tmp`. Worth noting SIGTERM (the ordinary CI-cancel signal) leaks the same way, since Python's default SIGTERM disposition skips `finally` too, so this is the common cancellation path rather than just the unblockable-SIGKILL corner.

Rather than the planned signal handler (which can't cover SIGKILL), consider not putting the token in the file at all: Maven interpolates `${env.NAME}` in settings.xml, so the generated file could carry `<password>${env.CLOUDSMITH_MAVEN_TOKEN}</password>` and the runner pass the token in the child's environment. A leftover file is then harmless.

BartoszBlizniak: Fable finding ^ left it as it's interesting for children projects, maven can be nasty...

cloudsmith_cli/credential_helpers/maven/installer.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+class MavenInstaller:
+    """Installs the Maven credential helper (shim + config entry)."""
+
+    name = "maven"
+    summary = "Maven credential helper for Cloudsmith repositories"
+    requires_repo = True
+
+    def install(
```
BartoszBlizniak, line 135: [`install`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/4a1cb9d4ad628e09f8ab118af4fedd812a02f16d/cloudsmith_cli/credential_helpers/maven/installer.py#L135) never checks that `--repo` actually exists, so a typo'd slug installs cleanly and only surfaces later as an opaque 401/404 from a wrapped `mvn` run. Since discovery already makes an authenticated API call at install time, a repo-existence check (or at least a warning when it can't be verified) would catch this where the user can still see the cause.

cloudsmith_cli/credential_helpers/maven/installer.py
```diff
@@ -0,0 +1,262 @@
...
+        actions.append(_deploy_snippet(binding))
+        return actions
+
+    def _configure(self, binding: config.Binding, description: str) -> list[str]:
+        """Persist the binding, write the shim, and warn about the setup."""
+        # The binding is persisted first: the shim intercepts every `mvn` on
+        # the machine and refuses to run one it has no binding for, so a shim
+        # written ahead of a failed set_binding would leave Maven unusable
+        # rather than merely uninstalled.
+        config.set_binding(binding)
```
BartoszBlizniak, line 189: [`config.set_binding(binding)`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/4a1cb9d4ad628e09f8ab118af4fedd812a02f16d/cloudsmith_cli/credential_helpers/maven/installer.py#L189) silently replaces any existing binding, and the binding is machine-wide - so installing for project B quietly re-points every `mvn` on the machine away from project A's repo. Probably fine as a v1 constraint, but the install output should say it *replaced* `old-org/old-repo` when it did, so the developer working across two projects understands what just changed.

cloudsmith_cli/cli/commands/credential_helper/shell.py
```diff
@@ -0,0 +1,55 @@
...
+_POSIX_STATEMENT = 'export PATH="{path}:$PATH"'
+
+_STATEMENTS = {
+    "bash": _POSIX_STATEMENT,
+    "zsh": _POSIX_STATEMENT,
+    "fish": 'fish_add_path "{path}"',
+}
+
+
+def detect_shell():
```
BartoszBlizniak, line 26: [`detect_shell`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/4a1cb9d4ad628e09f8ab118af4fedd812a02f16d/cloudsmith_cli/cli/commands/credential_helper/shell.py#L26) falls back to bash, so on Windows (where `$SHELL` is unset) `shell-init` prints POSIX `export PATH=...` that neither cmd nor PowerShell can eval - even though the launcher machinery happily writes an `mvn.cmd` shim there. If Windows isn't meant to be supported yet, it'd be better for `shell-init` to say so explicitly than to print a statement that can't work; otherwise a `powershell` entry in `_STATEMENTS` would close the gap.

cloudsmith_cli/cli/commands/exec_.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Maven runs use a generated ``settings.xml``; your ``~/.m2/settings.xml``
+    is not consulted. The repository comes from ``credential-helper install
+    maven``. The package manager is detected from the command, so just put it
+    after ``--``:
+
+    \b
+        $ cloudsmith exec -- mvn clean install
+    """
+    sys.exit(runner.run(list(command), credential=opts.credential))
```
BartoszBlizniak, line 34: [`exec`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/4a1cb9d4ad628e09f8ab118af4fedd812a02f16d/cloudsmith_cli/cli/commands/exec_.py#L34)'s help promises "a package-manager command authenticated against Cloudsmith", but an unsupported tool passes through with no credentials and no message - `cloudsmith exec -- npm ci` runs plain `npm` and the user has no signal that npm isn't wired up. A one-line stderr note ("npm is not a supported package manager; running unwrapped") would prevent people relying on authentication they aren't getting.

BartoszBlizniak: Fable find - might be worth  doublechecking docs vs execution

BartoszBlizniak (approved):
Pre-approving with some comments:

cloudsmith-iduffy:
> Pre-approving with some comments:

Moved this one to draft until we pull it into a cycle. I got carried away. 🙈

PR #341, opened 2026-08-03
feat(no-ticket): Add `cloudsmith tokens show` command 

BartoszBlizniak:
# Description

With the recent CLI changes, we are now forcing OIDC authentication via our integrations to be completed by the CLI. This has also removed the ability for customers to retrieve the short-lived authentication token, which could be used to authenticate with package managers that talk to Cloudsmith. 

This PR adds a new `show` sub-command to the `cloudsmith tokens` command, which exposes the currently authenticated token that is being used to talk with Cloudsmith.  

This PR restores the default GitHub Action behavior for `oidc-auth-only` without requiring additional code in the integration. 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ The CLI currently supports the following commands (and sub-commands):
 - `tokens`:               Manage API tokens.
   - `list`|`ls`:            List API tokens.
   - `refresh`:              Refresh an API token.
+  - `show`:                 Show the API token the CLI is authenticating with.
 - `upstream`:             Manage upstreams for a repository.
   - `cran`:                 Manage cran upstreams for a repository.
   - `dart`:                 Manage dart upstreams for a repository.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -352,6 +353,38 @@ For convenience the CLI will ask you if you want to install the default configur
...
+```bash
+TOKEN=$(cloudsmith tokens show)
+npm config set //npm.cloudsmith.io/example-org/example-repo/:_authToken "$TOKEN"
+```
+
+Use `--output-format json` to also see which credential source resolved the token and, for OIDC tokens, when it expires:
+
+```bash
+cloudsmith tokens show --output-format json
+{"data": {"auth_type": "api_key", "expires_at": "2026-07-31T12:34:56Z", "source": "oidc", "source_detail": "OIDC via GitHub Actions (org: example-org, service: example-service)", "token": "..."}}
```
cloudsmith-iduffy, line 385: We should add a version in here in the event we ever want to change the payload formatting. 

I wonder if we should merge this and https://github.com/cloudsmith-io/cloudsmith-cli/pull/338 both have similar goals.

BartoszBlizniak: Will pop this back into draft and re-do with #338 compatibility

BartoszBlizniak: well, we can just use 338 lol

PR #345, opened 2026-08-06
chore(ENG-13226): migrating from black/pylint to ruff

coillteoir:
Summary:
Updating CI to use ruff, and fixing lints from changes
Ensuring binaries workflow runs after code validation


cloudsmith-iduffy (approved):
The ruff check on pre-commit on CI fails now. Fix it up and you should be good for merging.

Issue #194, opened 2025-04-16
cloudsmith auth doesn't work in Cygwin bash shell


bakoontz2:
That's where python is installed, in Cygwin on Win 10.  Attempting to do a ```cloudsmith auth``` generates a link (which I can authenticate against) and this message:

```Starting webserver to begin authentication ...```

I'm assuming a web browser should pop up here.  I know that it's possible, because the Azure CLI does, in fact, pop up a webrowser window to authenticate against.

Is there maybe a command line option I'm missing?

nickxn:
hi @bakoontz2 - let me check, but I don't think we have an option like --no-browser that would allow you to workaround this. Have you tried using an API key, rather than SSO for auth?

BartoszBlizniak:
Hey @bakoontz2  / @nemith  - Thanks both for the feedback. 

We’ve added cloudsmith auth `--no-browser`, along with graceful fallback when the CLI cannot launch a browser automatically. For Cygwin and similar environments, the CLI now prints the SAML URL and continues waiting while you open it manually. This works when the browser and CLI are running on the same machine.

This does not yet fully support remote/headless sessions. The current SAML flow redirects to `localhost:12400`, which refers to the machine running the browser. If the CLI is running on a remote host, the callback cannot reach it without port forwarding.

This is still on our radar as an improvement to the CLI, we have some backend work to do and we are hoping to rehaul the auth on the back of that.

bakoontz2:
Awesome, appreciate this.  We've been restricted to using SSO for auth, so I don't really have a choice there (sorry, I missed your follow-up question...)

PR #350, opened 2026-08-13
fix(ENG-13962): degrade gracefully when auth can't open a browser (#194)

BartoszBlizniak:
# Description

Addresses the Cygwin and same-host headless-shell portion of issue #194. `cloudsmith auth` now handles both exceptions and falsey results from `webbrowser.open()` by showing a manual-open fallback and continuing to wait for the existing localhost SAML callback. A new `--no-browser` option on both `cloudsmith auth` and `cloudsmith authenticate` skips automatic browser launch and presents the IdP URL for the user to open manually. The URL banner is neutral in every mode, and existing informational output routing is preserved. This does not add support for remote or fully browserless authentication: the callback still binds to `127.0.0.1:12400`, and a device-authorization-style flow requires backend support outside this change.

The README's SAML section is updated for the new banner text, documents `--no-browser` and the automatic fallback, and now states the same-machine constraint on the callback. CHANGELOG has Added/Fixed entries under Unreleased.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

Regression coverage exercises the registered top-level Click command tree for `auth` and `authenticate`, including `webbrowser.Error`, an unexpected browser-launch exception, a false return value, `--no-browser`, successful automatic launch, and continuation into `AuthenticationWebServer.handle_request()`. Focused auth tests pass (14 passed), the full suite passes (632 passed, 40 skipped), and `pre-commit run --all-files` passes. Both command names expose `--no-browser` in generated local help. A live SAML round trip was not performed because no SSO-enabled test organization was available; browser-launch behavior and callback continuation are covered with local mocks, and no external service was modified.


cloudsmith-iduffy (approved):
Nice addition and makes sense.

PR #351, opened 2026-08-13
feat(ENG-56): derive Debian source package members from the .dsc on `push deb`

### LINKED ISSUE #56: RFE: 'cloudsmith push deb' parse .dsc `Files:` section for sources/changes
Currently to upload a debian source package:
```
cloudsmith push deb <repo/dist/version> foo_1.0.0-1.dsc --sources-file foo_1.0.0.orig.tar.bz2 --changes-file foo_1.0.0-1.debian.tar.xz
```
When scripting the CLI this means detecting/determining the sources/changes files (and they can have various suffix, sources: .orig.tar.gz, .orig.tar.bz2, changes: .debian.tar.xz, .diff.gz , etc).

The files that make up the source package are listed in the 'Files:' section of the .dsc file itself. https://manpages.debian.org/testing/dpkg-dev/dsc.5.en.html

It would be useful if 'cloudsmith push deb' when passed only a .dsc file could parse the 'Files:' to determine the additional sources/changes files it should upload.

If it helps, the https://pypi.org/project/pydpkg/ python library can parse .dsc files:
```
>>> dsc.source_files
['/tmp/testdeb_0.0.0.orig.tar.gz',
 '/tmp/testdeb_0.0.0-1.debian.tar.xz',
 '/tmp/testdeb_0.0.0.dsc' ]
 >>> dsc.all_files_present
True
```

BartoszBlizniak:
# Description

Closes #56.

`cloudsmith push deb` now works out a Debian source package's members from its `.dsc`, so this is enough:

```
cloudsmith push deb your-org/your-repo/debian/bookworm hello_2.10-3.dsc
```

where callers previously had to determine `--sources-file` and `--changes-file` themselves, across suffixes that vary per package (`.orig.tar.gz`, `.orig.tar.bz2`, `.debian.tar.xz`, `.diff.gz`, ...).

The `.dsc` is parsed plain or OpenPGP-clearsigned, from its `Checksums-Sha256:` field in preference to `Files:`. The upstream/native source archive becomes `sources_file` and the Debian packaging archive becomes `changes_file`, for the `1.0`, `2.0`, `3.0 (native)` and `3.0 (quilt)` source formats. Because uploading a Debian source package already means passing the `.dsc` as `PACKAGE_FILE` — the API rejects a binary `.deb` carrying source archives, and rejects a `.dsc` without them — the `.dsc` is parsed by default and no extra flag is needed. `--dsc-file` remains for naming a `.dsc` other than `PACKAGE_FILE`, and an explicit `--sources-file` or `--changes-file` still wins for its own field. Passing `--sources-file` skips derivation entirely, which keeps an escape hatch for anything the parser will not accept.

Members must be bare filenames resolving to regular files next to the `.dsc`; absolute paths and directory traversal in the `.dsc` are rejected, and each member is uploaded by its canonical path so a symlink cannot be swapped between the check and the read. Symlinked members are allowed, since `mk-origtargz --symlink` (the `uscan` default) routinely links the `.orig` tarball in from a download cache.

Following the discussion on the issue, a detached upstream signature (`*.orig.tar.*.asc`) is skipped with a warning — the deb package format has no field to carry it, and nothing in the uploaded source is lost — while a multi-component source package (`*.orig-<component>.tar.*`) is rejected, because dropping a component would upload incomplete source. `3.0 (git)`, `3.0 (bzr)` and `3.0 (custom)` are reported as unsupported source formats rather than being half-handled.

The implementation stays a small `deb`-specific layer over the API-model-driven push command generation. Other formats and the generic upload flow are unchanged.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

Verified end-to-end against a live repository with three real source packages: Debian's own `hello_2.10-3` (`3.0 (quilt)`, including the `.orig.tar.gz.asc` it ships), a `3.0 (native)` package, and a `1.0` non-native package with a `.diff.gz`. All three uploaded and synchronised, with the `.dsc` and its archives present on the resulting package.

Regression coverage invokes the registered `push deb` command tree and asserts both the local paths seen at validation and the uploaded identifiers seen at package creation, for native, quilt and legacy non-native layouts; plus `.dsc`-as-`PACKAGE_FILE` derivation, a binary `PACKAGE_FILE` never being parsed, explicit-option precedence in both directions, OpenPGP-clearsigned control data, absolute paths, traversal, symlinked members, multi-component rejection, skipped detached signatures, unsupported source formats, names and versions containing `.orig-`, malformed file-list rows, conflicting checksum lists, and `deb`-only option registration.


cloudsmith-iduffy (approved):
If we know of someone internally that knows the debian package format well it would be good to get them to give this a glance. Otherwise looks fine to me.

PR #353, opened 2026-08-14
fix: embedded additional keyrings for headless use cases

cloudsmith-iduffy:
# Description

When using the single binary it is not possible to install different keyring plugins. Some users may want to use cloudsmith cli in a headless environment but still use their sso token.

In the previous scenario where a user installed cloudsmith cli into a python environment they could install something like keyrings.cryptfile or keyrings.alt and be able to store their sso generated credentials to a file instead of the operating system keychain. They could then mount that file in a docker container or share it in ssh sessions.

By building keyrings.cryptfile and keyrings.alt we enable users of the CLI to use the keyring environment variables to manipulate the keyring that is loaded and thus customise where cloudsmith stores credentials. 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


BartoszBlizniak (approved):
Pre-approving, but can we please check manually once on the produced standalone binary that it all works correctly (if haven't already)

cloudsmith-iduffy:
> Pre-approving, but can we please check manually once on the produced standalone binary that it all works correctly (if haven't already)

Yeah have tested it locally and will test it again when the binaries get built.

cloudsmith-iduffy:
Tested the binaries CI built and all worked as expected.

PR #356, opened 2026-08-14
fix: add CLOUDSMITH_KEYRING_* env var aliases to avoid global impact

cloudsmith-iduffy:
## Summary

- Adds `CLOUDSMITH_KEYRING_BACKEND` as an alias for `PYTHON_KEYRING_BACKEND`, and `CLOUDSMITH_KEYRING_KEY` as an alias for `KEYRING_PROPERTY_KEYRING_KEY`
- Aliases only apply when the native env var isn't already set, so existing `PYTHON_KEYRING_BACKEND`/`KEYRING_PROPERTY_KEYRING_KEY` usage is unaffected
- Applied consistently across all three keyring entry points (`_get_value`, `_set_value`, `_delete_value`)

## Motivation

Ran into this trying to use [cloudsmith-keyring](https://github.com/cloudsmith-labs/cloudsmith-keyring) in an existing project via the standalone CLI binary. Setting `PYTHON_KEYRING_BACKEND` in the environment so the CLI would pick it up also got picked up by unrelated Python tooling in the same environment — `uv sync` started failing because of it. These aliases let the backend/key be configured for the CLI specifically, without a global env var affecting every other Python process.

## Test plan

- [x] Added unit tests covering: alias sets the var when unset, alias does not override an existing native var, no-op when alias is unset, and that resolved backend properties are applied via `set_properties_from_env()`
- [x] `pytest cloudsmith_cli/core/tests/test_keyring.py` — 42 passed
- [x] pre-commit hooks (ruff format/check, etc.) pass


cloudsmith-iduffy:
@claude review once

BartoszBlizniak (approved):

cloudsmith-iduffy:

PR #369, opened 2026-08-20
feat(ENG-14032): adding container image build to CI

coillteoir:


Dockerfile
```diff
@@ -1,44 +1,46 @@
...
-      arm64) CS_ARCH="aarch64" ;; \
-      *) echo "Unsupported architecture: ${TARGETARCH}" >&2; exit 1 ;; \
-    esac; \
-    ARCHIVE="cloudsmith-${CLOUDSMITH_CLI_VERSION}-linux-${CS_ARCH}-musl.tar.gz"; \
-    cd /tmp/binaries; \
-    sha256sum -c "${ARCHIVE}.sha256"; \
-    mkdir -p /opt; \
-    tar -xzf "${ARCHIVE}" -C /opt
+
+ARG TARGETARCH=x86_64
```
cloudsmith-iduffy: ideally we don't default this it should be coming from the container runtime https://docs.docker.com/build/building/multi-platform/#:~:text=%2C%20TARGETOS%20and-,TARGETARCH,-build%20arguments.

Dockerfile
```diff
@@ -1,44 +1,46 @@
...
 
-COPY binaries/ /tmp/binaries/
+ADD bin bin
+ADD cloudsmith_cli cloudsmith_cli
+ADD packaging packaging
+ADD pyproject.toml pyproject.toml
+ADD uv.lock uv.lock
+ADD VERSION VERSION
 
-RUN set -eu; \
```
cloudsmith-iduffy, line 10: do we still maintain the multi-architecture images by changing this?

Dockerfile
```diff
@@ -1,44 +1,46 @@
...
-    cd /tmp/binaries; \
-    sha256sum -c "${ARCHIVE}.sha256"; \
-    mkdir -p /opt; \
-    tar -xzf "${ARCHIVE}" -C /opt
+
+ARG TARGETARCH=x86_64
+ARG CLOUDSMITH_CLI_VERSION
+ARG PYTHON_VERSION=3.14
+
+RUN uv sync --locked --no-dev --no-editable --group binary --extra all --python "${PYTHON_VERSION}"
```
cloudsmith-iduffy: previously we had no python or uv dependency due to just dropping in the standalone binary that takes care of all those concerns.

Dockerfile
coillteoir: I can add a matrix to build it for `amd64` and `aarch64` if needed.

Dockerfile
coillteoir: Fair, I don't think this variable is referenced anyways. Some muscle memory from multiarch go builds

Dockerfile
coillteoir: The final image in this build only contains the standalone binary. 

It is ideal to build container images from source, since it means builds are more declarative (docker isn't a declarative build format since it calls out to the internet, but its better than copying from local directories.)

Dockerfile
BartoszBlizniak: more the merrier 😛 our release workflow pushes the docker image to [cloudsmith](https://broadcasts.cloudsmith.com/cloudsmith/cli) and [dockerhub](https://hub.docker.com/r/cloudsmith/cloudsmith-cli)

Dockerfile
```diff
@@ -1,44 +1,44 @@
 ARG ALPINE_IMAGE=alpine:3.21@sha256:48b0309ca019d89d40f670aa1bc06e426dc0931948452e8491e3d65087abc07d
 
-FROM ${ALPINE_IMAGE} AS unpack
+FROM astral/uv:alpine AS build
```
BartoszBlizniak: Could we keep the production image consuming the [checksummed, smoke-tested musl artifacts](https://github.com/cloudsmith-io/cloudsmith-cli/blob/6a3b6b20f0630f25aeaff91e913414dc0e2280f9/.github/workflows/binaries.yml#L129-L167) and validate that target for both architectures? This instead [rebuilds with floating `astral/uv:alpine` and Python 3.14](https://github.com/cloudsmith-io/cloudsmith-cli/blob/6a3b6b20f0630f25aeaff91e913414dc0e2280f9/Dockerfile#L3-L20), rather than using the [pinned Python 3.12 binary pipeline](https://github.com/cloudsmith-io/cloudsmith-cli/blob/6a3b6b20f0630f25aeaff91e913414dc0e2280f9/.github/workflows/binaries.yml#L82-L105). The [new PR job runs only on amd64](https://github.com/cloudsmith-io/cloudsmith-cli/blob/6a3b6b20f0630f25aeaff91e913414dc0e2280f9/.github/workflows/image-build.yml#L12-L25), so this arm64 path first runs under QEMU during a release and is pushed without a smoke test. A pinned source-build target could remain as an additional fast PR check.

.github/workflows/release.yml
```diff
@@ -452,24 +452,6 @@ jobs:
...
-          merge-multiple: true
-          path: binaries
-
-      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-        with:
-          pattern: cloudsmith-linux-x86_64-gnu
-          merge-multiple: true
-          path: binaries
-
-      - name: Set up the built cloudsmith binary on the runner
```
BartoszBlizniak, line 467: This removes the [artifact download and setup step](https://github.com/cloudsmith-io/cloudsmith-cli/blob/94f9cfefeb9b1db6bc3c2f96a1f5eb316b498439/.github/workflows/release.yml#L455-L471) that installs `cloudsmith`, but the job still [runs `cloudsmith whoami` and installs the credential helper](https://github.com/cloudsmith-io/cloudsmith-cli/blob/6a3b6b20f0630f25aeaff91e913414dc0e2280f9/.github/workflows/release.yml#L459-L463). On a fresh release runner, won't this fail with `cloudsmith: command not found`? I think we still need the GNU artifact download and [`setup-cloudsmith-binary`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/6a3b6b20f0630f25aeaff91e913414dc0e2280f9/.github/actions/setup-cloudsmith-binary/action.yml) with `add-local-bin: "true"`.

cloudsmith-iduffy (approved):

PR #371, opened 2026-08-21
feat(ENG-13678): adding cargo credential helper to cli

coillteoir:


tigh-latte (commented):
v good stuff, I've spotted nothing that copilot hasn't already

coillteoir:
> v good stuff, I've spotted nothing that copilot hasn't already

Thanks! Working through copilot issues, and will run another pass

cloudsmith-iduffy (approved):
Looks good! took it for a spin on one of our internal rust projects and credentials resolved as expected. I kicked off https://github.com/cloudsmith-io/cloudsmith-cli/actions/runs/32853891342 on your branch to have publicly accessible binaries to see how the helper would get on in some pre-existing CI flow.

cloudsmith_cli/core/cache_utils.py
```diff
@@ -109,14 +114,22 @@ def merge_json_file(
...
-
+        match format:
+            case "json":
+                try:
+                    parsed = json.loads(existing_text)
+                except (json.JSONDecodeError, ValueError):
+                    pass
+            case "toml":
+                try:
+                    parsed = toml.loads(existing_text)
```
cloudsmith-iduffy: I believe this might result in comments being stripped from the toml file when we do the dumps down further. 

https://pypi.org/project/tomlkit/ will preserve everything.

coillteoir: Updated with tomlkit

PR #383, opened 2026-08-23
fix: stop recurring macOS keychain prompts and self-heal dead SSO sessions

cloudsmith-iduffy:
# Description

## Motivation

On macOS, every SSO session produced a stream of "cloudsmith wants to use your credential information" prompts. Clicking "Always Allow" did not stop them. The prompts got worse for anyone who runs the CLI from more than one install (standalone binary, a uv environment, `uv run`).

The root cause is in the `keyring` library's macOS backend: every write is a delete followed by a re-create (`SecItemDelete` + `SecItemAdd`). A re-created item is a new keychain item with a fresh access control list, so it forgets every "Always Allow" the user ever granted. The CLI rewrites its three token items on every refresh — at most every 30 minutes — so grants never survived, and each install flavor kept stealing the items from the others.

Two related problems surfaced while fixing this:

1. **Keyring entries were not profile-scoped.** Profiles that target the same API host shared one set of SSO tokens and clobbered each other.
2. **Dead tokens got stuck.** When the server rejected a refresh (for example after a lost refresh-token rotation), the CLI kept the dead tokens, warned every 30 minutes, and served a stale bearer token in between. There was no way back to a working state short of a manual logout.

## Changes

**Keychain writes now update items in place.** A new `core/macos_keychain.py` binds `SecItemUpdate` via ctypes. `_set_value` tries it first on macOS and falls back to the normal `keyring` write when the item does not exist. `SecItemUpdate` changes the stored secret and keeps the item's access control list, so "Always Allow" now sticks permanently. Each binary pays its grants once, ever. Note: `keyring.get_keyring()` returns a chainer backend that delegates writes to its first member, so the backend gate inspects `backends[0]`, not the chainer.

**Keyring entries are profile-scoped.** Until now, every profile stored its SSO tokens under one unscoped service name per API host (for example `cloudsmith_cli-access_token-<host>`). This change appends `-profile-<name>` to the service name for non-default profiles, giving each profile its own token set. The old unscoped entries are what the code calls **legacy entries**, and existing sessions live in them, so backwards compatibility works in two parts:

- The default profile keeps using the unscoped names. Nothing changes for users who never pass a profile — the same entries are read and written as before.
- A non-default profile first reads its own scoped entries and falls back to the unscoped ones when they do not exist. An existing session therefore keeps working after the upgrade; the first successful refresh writes the rotated tokens to the scoped entries, completing the migration without a re-login. Logout for a scoped profile removes both its scoped entries and the unscoped ones, so no stale tokens are left behind.

**Rejected refreshes self-heal.** When the refresh endpoint definitively rejects the stored tokens (400, 401, 403, or 422 — the API returns 422 for an invalid refresh token), the CLI deletes that profile's entries and returns to a clean logged-out state with a clear "run `cloudsmith auth`" message. The wipe touches only the profile's own entries; if the profile was still running on the unscoped entries via the fallback, those hold the dead tokens and are removed instead — so one profile's rejection can never destroy another profile's session. Transient failures (network errors, 5xx) keep the existing behavior: stamp the attempt time and retry after the 30-minute throttle. A refresh response without an access token is treated as a failure instead of producing an `Authorization: Bearer None` header, and a locally missing refresh token skips the refresh instead of posting an empty token.

## Verification

Verified live on macOS against a real API host: three consecutive forced refresh cycles rotated the access token each time while all three keychain items kept their original creation date — same items, updated in place, access grants intact. The item-identity check is `security find-generic-password -s <service> | grep cdat`; a changing `cdat` means the item was re-created and the ACL was reset. The rejection path was also verified live: a dead refresh token produced a 422, and the CLI wiped the entries and dropped cleanly to logged-out. There are no macOS-only automated tests since CI has no macOS runners; the unit tests are cross-platform.

## Type of Change

- [x] Bug fix

## Additional Notes

**Phase-out plan for the legacy unscoped entries.** The unscoped service names stay as the default profile's canonical storage — they are only "legacy" from the point of view of non-default profiles. The fallback exists to migrate existing sessions without a forced re-login, and it retires itself:

1. **This release.** A non-default profile reads the unscoped entries only when its own scoped entries do not exist. The first successful refresh writes the tokens to the scoped entries, so the fallback becomes dead code for that profile within 30 minutes of normal use. A logout or a rejected refresh also removes the unscoped entries the profile was running on.
2. **Some time in the future.** Remove the fallback read (`_get_value_with_fallback`) and the `include_legacy` handling in `delete_sso_tokens`. Any session that never migrated — a user who skipped the intermediate releases — resolves no credentials and gets the standard "run `cloudsmith auth`" message, not an error.
3. The window can stay short: profile use is effectively limited to Cloudsmith employees, and active sessions migrate automatically on their first refresh after upgrading.


cloudsmith-iduffy:

cloudsmith_cli/core/tests/test_keyring_provider.py
BartoszBlizniak, line 68: **Blocking:** `test_returns_none_on_refresh_failure` patches `saml.refresh_access_token`, but the provider now binds `refresh_access_token` at module level (line 7), so the patch no longer intercepts. Two consequences:

- On Linux CI, the real `refresh_access_token` runs against the `MagicMock` session, "succeeds" with mock tokens, and the test passes without ever exercising the `ApiException(401)` path it claims to test.
- On a macOS dev machine, the mock tokens flow into the real `store_sso_tokens` → `_update_keychain_item_in_place` (this file has no `get_keyring` patch, so the real macOS backend is seen) → `update_generic_password(..., value=MagicMock)`, where ctypes' `_as_parameter_` resolution on a MagicMock recurses at C level and **segfaults the whole pytest run**. Had the value been a plain string, the test would have written into the developer's real keychain.

Fix: patch `keyring_provider.refresh_access_token` like the new tests do, and also patch `keyring.delete_sso_tokens` (with the corrected target, the 401 triggers the wipe path, which would otherwise really delete keychain entries on a dev Mac). Consider an autouse `keyring.get_keyring` patch for this file, mirroring `test_keyring.py`, so no test here can ever reach a real backend.

cloudsmith_cli/core/credentials/providers/keyring_provider.py
```diff
@@ -4,12 +4,38 @@
...
 
+from ....cli.saml import refresh_access_token
 from ....core import keyring
+from ...api.exceptions import ApiException
 from ..models import CredentialContext, CredentialResult
 from ..provider import CredentialProvider
 
 logger = logging.getLogger(__name__)
 
+REFRESH_REJECTED_STATUSES = (400, 401, 403, 422)
```
BartoszBlizniak, line 15: Corporate proxies and WAFs commonly return 403 (and sometimes 400) for reasons unrelated to token validity, and that would wipe a valid session here. The PR description says the API returns 422 for an invalid refresh token, and 401 is the other unambiguous signal - is there a case where the API itself returns 400/403 for a dead token? If not, `(401, 422)` would avoid wiping sessions on WAF/proxy interference. The cost of a false positive is only a forced re-auth, so this is a judgment call - but worth making deliberately.

cloudsmith_cli/cli/commands/logout.py
```diff
@@ -43,15 +43,15 @@ def _clear_keyring(api_host, dry_run, use_stderr):
...
+    if not keyring.has_sso_tokens(api_host, profile=profile):
         click.echo("No SSO tokens found in system keyring.", err=use_stderr)
         return {"action": "not_found"}
 
     if dry_run:
         click.echo("Would remove SSO tokens from system keyring.", err=use_stderr)
         return {"action": "would_remove"}
 
-    deleted = keyring.delete_sso_tokens(api_host)
+    deleted = keyring.delete_sso_tokens(api_host, profile=profile)
```
BartoszBlizniak: `delete_sso_tokens(api_host, profile=profile)` defaults to `include_legacy=True`, so `cloudsmith logout --profile staging` also deletes the unscoped entries - which are the default profile's *permanent canonical storage*, not just pre-migration leftovers. A user with an active default-profile session who logs out of a secondary profile loses both sessions.

The refresh-rejection path already gets this right (`_handle_refresh_failure` only falls back to the legacy delete when the profile has no scoped entries). Logout could mirror that: delete legacy entries only when the profile's scoped entries don't exist, i.e. when the profile was still running on the fallback.

cloudsmith_cli/core/credentials/providers/keyring_provider.py
```diff
@@ -4,12 +4,38 @@
 
 import logging
 
+from ....cli.saml import refresh_access_token
```
BartoszBlizniak: Moving this import to module level makes `core` depend on `cli` at import time, which is what forced the deferred provider imports in `chain.py`. `refresh_access_token` (and `raise_for_api_error`) only use `requests` and `core.api.exceptions.ApiException` - no click, no CLI state - so they could move into `core` (e.g. `core/sso.py` or `core/api/`), letting `cli/saml.py` re-export them. That deletes the circular-import workaround instead of managing it.

BartoszBlizniak (changes requested):
few comments before approving

cloudsmith_cli/cli/commands/logout.py
cloudsmith-iduffy: Oh interesting! good catch, let me investigate that. I was trying to get it so logout cleans up legacy entries so it having a profile specified and clearing out the non-profiled one is a bug.

cloudsmith_cli/core/credentials/providers/keyring_provider.py
cloudsmith-iduffy: Going to leave this one under the assumption users have direct access to our APIs.

cloudsmith_cli/core/credentials/providers/keyring_provider.py
cloudsmith-iduffy: Fixed!

cloudsmith_cli/core/tests/test_keyring_provider.py
cloudsmith-iduffy: Fixed!

BartoszBlizniak (approved):

PR #389, opened 2026-08-25
feat(no-ticket): add `cloudsmith repos privileges` command group

BartoszBlizniak:
# Description

Adds a `cloudsmith repos privileges` command group for managing who has explicit access to a repository: teams, users and service accounts, all treated the same way.

Four subcommands, split by what they can take away:

| Command | Method | Effect on access | Prompts |
| --- | --- | --- | --- |
| `list` | GET | Read-only | no |
| `set` | PATCH | Adds or raises only, anything unnamed is untouched | no |
| `revoke` | PUT | Revokes the named teams, users and services | yes |
| `replace` | PUT | The file becomes the whole truth, anything absent is revoked | yes |

Only `revoke` and `replace` can remove access, so only those two confirm before writing. Both take `-y`.

```
$ cloudsmith repos privileges set your-org/your-repo \
    --team your-team --service your-ci --privilege write

Granting Write on your-repo in the your-org namespace to team your-team, service your-ci ... OK

Type    | Name      | Privilege
Service | your-ci   | Write
Team    | your-team | Write

Results: 2 privileges
```

```
$ cloudsmith repos privileges revoke your-org/your-repo --team your-team --user someone-else

Getting list of repository privileges ... OK
No explicit privilege for user someone-else, skipping.
Revoke the privileges of team your-team on your-repo in the your-org namespace? [y/N]: y
Revoking the privileges of team your-team ... OK

Type | Name       | Privilege
User | their-user | Read

Results: 1 privilege
```

```
$ cloudsmith repos privileges list your-org/your-repo -F json \
    | jq '.data' \
    | cloudsmith repos privileges replace your-org/your-repo - -y
```

Three API constraints shaped the design, and are worth knowing before reviewing:

- **No paging.** The endpoint returns every privilege in one response and ignores page parameters, so `list` deliberately offers no page options rather than advertising paging that never happens.
- **No single delete.** There is no way to delete one privilege, so `revoke` reads the current set and writes back what is being kept. A concurrent change by someone else can therefore be lost, which the command's help says out loud.
- **One target per privilege.** A privilege names a team *or* a user *or* a service, never two. That is why the table reads `Type | Name | Privilege` instead of a sparse column per kind, and why both `set` and `replace` reject an entry that names more than one.

Two things outside the command module:

- `cli/utils.py` `confirm_operation()` now accepts an explicit empty `prefix` so a command can ask its question directly instead of through the shared "Are you absolutely certain you want to..." preamble. Existing callers pass no prefix and are unaffected.
- `cli/exceptions.py` `handle_api_exceptions()` takes an optional `summarise_error` callable, which receives the exception and can decline by returning `None`. The privileges commands use it to turn the API's field-indexed 422 into one sentence and decline everything else, so a 403, a 404, or a 422 the API declines to explain all keep their status code and hint. Every other caller keeps the existing block unchanged.

```
# Before
Failed to set the repository privileges! (status: 422 - Unprocessable Entity)

Detail: Invalid input.
Privileges Field: Invalid team(s) specified ['no-such-team']

# After
Could not set privileges for your-repo: invalid team(s) specified ['no-such-team']
```

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

Exercised end to end against a live organisation on a scratch repository, which is where the one real bug turned up: a listed privilege carries an explicit `null` for the two target kinds that do not apply, and the PUT endpoint rejects those nulls outright, so `revoke` failed on its write-back. Writes now send only the key that applies. Unit tests cover that shape directly, since a mocked response that omits the nulls would not have caught it.

Also verified by hand, because the unit tests only assert the shape and not the API's actual behaviour: `set` really is an upsert (naming one team leaves every other privilege alone), running `revoke` twice is a no-op that still exits 0, and `list -F json | jq '.data' | replace - -y` round-trips because listing and replacing share an entry shape.

`set` does not report that a named target already had the privilege it was given. The PATCH returns 204 with no after-state, so saying "already had this" would mean a GET before every write to produce an advisory line nobody acts on. It prints the entries it wrote rather than the repository's resulting state; `list` is there for that.

Two edges worth a reviewer's attention, both found by running this against a live org rather than by the tests:

- `replace` with a file that lists nothing revokes every explicit privilege, and that includes the caller's own. On a private repository that means losing sight of it entirely, so the confirmation for an empty file drops the "replace all 0 privileges" phrasing and asks `Revoke all explicit access to your-repo in the your-org namespace?` instead.
- `replace your-org/your-repo -` reads the document from stdin, which leaves nothing for the confirmation to read an answer from. It now refuses up front and asks for `-y` rather than aborting with no explanation.

`revoke` has to write the whole list back, so every privilege it keeps has to be one it can express. If it reads an entry it cannot (no principal, or a kind a future API adds), it refuses and points at `replace` rather than silently revoking it as a side effect. Declining either confirmation writes nothing and prints nothing, including under `-F json`; the paths that run to completion always emit a document.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Added `cloudsmith repos privileges` for managing explicit repository access from the terminal. `list` shows the teams, users and service accounts that were granted access explicitly; `set` grants or raises access for any number of them and leaves everyone else untouched; `revoke` takes access away from the ones named, skipping any that had none; and `replace` makes a JSON file (or stdin) the complete truth for the repository. `revoke` and `replace` ask for confirmation first unless `-y` is passed.
+
 ## [1.25.0] - 2026-08-24
 
 ### Added
--- a/cloudsmith_cli/cli/commands/repos.py
+++ b/cloudsmith_cli/cli/commands/repos.py
@@ -60,6 +60,65 @@ def print_repositories(opts, data, page_info=None, show_list_info=True, page_all
     )
 
 
+PRIVILEGE_LEVELS = ("read", "write", "admin")
+
+# A privilege names a team, a user or a service account, never more than one,
+# which is why the table has a Type/Name pair rather than a column per kind.
+TARGET_KINDS = ("team", "user", "service")
+
+
+def get_privilege_target(entry):
+    """Get the (kind, name) pair a privilege entry applies to, if any."""
+    for kind in TARGET_KINDS:
+        name = entry.get(kind)
+        if name:
+            return kind, name
+    return None
+
+
+def as_privilege_entry(target, privilege):
+    """Build the compact entry shape the API accepts on write.
+
+    A listed privilege carries every target key, with null for the two that
+    don't apply, and the write endpoints reject those nulls outright.
+    """
+    kind, name = target
+    return {"privilege": privilege, kind: name}
+
+
+def print_privileges(opts, data, show_list_info=True):
+    """Print repository privileges as a table or output in another format."""
+    headers = ["Type", "Name", "Privilege"]
+
+    targeted = [(get_privilege_target(entry), entry) for entry in data]
+    rows = []
+    for target, entry in sorted(targeted, key=lambda item: item[0] or ("", "")):
+        kind, name = target or ("", "")
+        rows.append(
+            [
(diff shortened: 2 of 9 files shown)


BartoszBlizniak:
Superseded by #393 - the head branch was renamed to `eng-14008-cli-add-repo-privileges-command`, which closed this one. Same commit, same content.

PR #390, opened 2026-08-25
feat(issue-83): add `cloudsmith repos gpg` command group

BartoszBlizniak:
# Description

Adds `cloudsmith repos gpg` for managing the GPG key a repository signs its package indexes with, and applies the command-design review of it.

`get` shows the active key and its armored public block. `upload` installs a key you supply. `regenerate` replaces the current key with a freshly generated Cloudsmith one. There is no `delete` - the API has no way to remove a repository's key, so offering one would imply a capability that doesn't exist.

```
$ cloudsmith repos gpg get your-org/your-repo
Getting GPG key ... OK

Fingerprint: 383E633DBCDB51EBD4DF418103BD5E7D97AFEA56
Fingerprint (short): 03BD5E7D97AFEA56
Active: True
Default: True
Comment: your-org/your-repo

Public Key:
-----BEGIN PGP PUBLIC KEY BLOCK-----
...
```

Three things in here are worth a reviewer's attention, because they are judgement calls rather than plumbing.

**Regenerating asks for a typed word, not a y/N.** It discards the old key with no undo, and every consumer pinned to the old fingerprint has to pick up the new one before their next install succeeds, so a reflexive `y` is too cheap:

```
$ cloudsmith repos gpg regenerate your-org/your-repo

Regenerating a repository's GPG key is irrevocable. The old key is discarded
and every consumer verifying against its fingerprint will need to fetch and
trust the new one before their next install succeeds.

Type 'regenerate' to confirm: regenerate
Regenerating GPG key ... OK
```

Anything else typed declines with `Not confirmed. No changes made.` and sends nothing. With no terminal attached the command fails with a usage error instead of blocking on a question nobody can answer - `-y/--yes` remains the way to run it in a pipeline.

**Both mutating subcommands take `-n/--dry-run`.** It validates the local inputs - an empty key file still errors - then reads the key currently in place and names the fingerprint the real run would replace, so a mistyped repository or an expired credential fails in the rehearsal rather than on the attempt. It stops before the mutating request, and it never asks for a passphrase, since nothing is being sent.

```
$ cloudsmith repos gpg regenerate your-org/your-repo --dry-run
Would regenerate the GPG key for your-repo in the your-org namespace, replacing 41028EE9E834351ECFE1012984C2EFC769E5D98A. Nothing sent - this was a dry run.

$ cloudsmith repos gpg regenerate your-org/typo --dry-run
ERROR
Could not regenerate GPG key for typo: not found.
```

**The actionable failures read as one sentence.** `handle_api_exceptions` gained an optional per-status summary map, which the GPG commands use for the three statuses a person can act on:

```
# Before
Failed to set the repository GPG key! (status: 402 - Payment Required)

Detail: Custom GPG keys are not active; upgrade your account!

# After
Could not set GPG key for your-repo: custom GPG keys require a paid plan.
```

Any other status, and JSON output in every case, keeps the existing context/detail/hint rendering and the full error envelope - so no other command's output changes.

`--private-key-file -` reads the key from stdin, which means the passphrase cannot also come from stdin - stdin has already been drained. That conflict is rejected up front, and the rejection is keyed off the literal `-` recorded at parse time rather than off the identity of the returned stream: `click.File` hands back a fresh stream object for `-` on every conversion, so an `is`-comparison against `click.get_text_stream("stdin")` holds under `CliRunner` but never in a real process, which would have left the check inert exactly where it matters.

Secret handling is deliberately narrow: key material and passphrases are only ever read from a file, stdin, or a hidden prompt, never from a command-line value that would land in shell history and the process list. `--debug` is refused on `upload`, because debug mode logs the raw request. The passphrase prompt only appears when a terminal is attached; without one the key is taken to be unencrypted rather than aborting on EOF, and under `-F json` the prompt goes to stderr so stdout stays a single parseable document.

## Type of Change

- [x] New feature
- [x] Bug fix (test isolation - see below)
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

One fix in here is unrelated to GPG and worth flagging: `config.get_or_create_options` caches the Options object in a thread-local, so state that is sticky by design (`--debug`) leaked from one test's CLI invocation into every later one in the same process. Three tests this branch doesn't touch (two in `test_whoami.py`, one in `test_exceptions.py`) fail without it, and any test running after one that passes `--debug` was exposed - which test depends on ordering. An autouse fixture now clears the thread-local for every CLI test. Nothing about production behaviour changes - a real invocation only ever builds Options once.

Verified against a live organisation on a throwaway repository (since deleted): reading the key, the typed confirmation both accepted and declined, a real `regenerate` producing a new fingerprint, both dry-runs, the non-terminal refusal, and the 400 and 404 error voices. The interactive confirmation was driven over a pty; `CliRunner` has no terminal, so the tests that exercise the prompt patch the terminal check explicitly.

**N/A:** no screenshots - this is a CLI change, and the terminal output above covers it.


BartoszBlizniak:
Superseded by #391 - the branch was renamed to `eng-14007-cli-add-gpg-key-command` and GitHub closed this PR rather than retargeting it. Same commits, same content.

PR #391, opened 2026-08-25
feat(ENG-14007): add `cloudsmith repos gpg` command group

BartoszBlizniak:
# Description

Adds `cloudsmith repos gpg` for managing the GPG key a repository signs its package indexes with: `get` reads the active key, `upload` installs one you supply, `regenerate` replaces it with a fresh Cloudsmith key. There's no `delete` - the API has no way to remove a repository's key.

Three judgement calls worth a look; the rest is plumbing.

**`regenerate` needs a typed word, not a `y/N`.** It's irreversible and breaks every consumer pinned to the old fingerprint, so a reflexive `y` is too cheap. With no terminal attached it fails instead of blocking; `-y` is still how a pipeline runs it.

```
$ cloudsmith repos gpg regenerate your-org/your-repo

Regenerating a repository's GPG key is irrevocable. The old key is discarded
and every consumer verifying against its fingerprint will need to fetch and
trust the new one before their next install succeeds.

Type 'regenerate' to confirm: regenerate
Regenerating GPG key ... OK
```

**`upload` and `regenerate` take `-n/--dry-run`.** It validates the inputs, reads the key currently in place and names the fingerprint that would be replaced, then stops - so a mistyped repository or a dead credential fails in the rehearsal rather than on the attempt. It never prompts for a passphrase, since nothing is sent.

**Actionable API failures read as one sentence.** `handle_api_exceptions` gained an optional per-status summary map, used for the three statuses a person can act on: `Could not set GPG key for your-repo: custom GPG keys require a paid plan.` Every other status, and all JSON output, keeps the existing rendering - no other command's output changes.

Secrets only ever come from a file, stdin or a hidden prompt, never a command-line value. `--debug` is refused on `upload`, because it would log the request body.

## Type of Change

- [x] New feature
- [x] Bug fix (test isolation - see below)
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Other (please describe)

## Additional Notes

**Two changes here aren't about GPG.** Both were found while building this, and both explain files outside `repos.py` in the diff:

1. `cli/tests/conftest.py` - `config.get_or_create_options` caches Options in a thread-local, so sticky state (`--debug`) leaked between CLI invocations in the same test process. Three tests this branch doesn't touch fail without the new autouse fixture. Test-only: a real invocation builds Options once.
2. `cli/exceptions.py` - the standard error block moved into a helper so the summary map could sit beside it. Behaviour-identical for every other caller.

Verified against a live organisation on a throwaway repository (since deleted): the typed confirmation accepted and declined, a real `regenerate`, a real key upload, both dry runs, the non-terminal refusal, and the 400/404 error voices. The prompt was driven over a pty - `CliRunner` has no terminal, so the tests that exercise it patch the terminal check.

**N/A:** no screenshots - CLI change, and the output above covers it.


cloudsmith-iduffy (approved):

BartoszBlizniak:

PR #396, opened 2026-08-26
fix(no-ticket): Prefix pnpm credential with Bearer

tigh-latte:
# Description

pnpm makes a decision based on some regex on whether or not to prefix the token generated by its `tokenHelper` output with `Bearer `.

For some reason, this regex is not being tripped up when ran in certain scenarios for me (inside a docker container for one), so, always prefix with Bearer instead.

## Type of change
- [x] Bug fix

### DIFF
--- a/cloudsmith_cli/credential_helpers/pnpm/runtime.py
+++ b/cloudsmith_cli/credential_helpers/pnpm/runtime.py
@@ -42,7 +42,7 @@ def get_pnpm_credentials(server_url, credential=None, api_host=None, org=None):
     ):
         return None
 
-    return credential.api_key
+    return "Bearer " + credential.api_key
 
 
 def execute(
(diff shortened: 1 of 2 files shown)


coillteoir (approved):

cloudsmith_cli/credential_helpers/pnpm/runtime.py
```diff
@@ -42,7 +42,7 @@ def get_pnpm_credentials(server_url, credential=None, api_host=None, org=None):
     ):
         return None
 
-    return credential.api_key
+    return "Bearer " + credential.api_key
```
cloudsmith-iduffy: We may want to resolve this via the credential auth_type field. I believe if this is a OIDC jwt or a SAML jwt then its `bearer` but if its a straight up API key then its `token`. https://github.com/cloudsmith-io/cloudsmith-cli/pull/394/changes#diff-f72e9d51919dea0f9c9dda03820678db0e03847208938a0efda15232c94ac088R123

tigh-latte: ah nice spot

PR #400, opened 2026-08-31
feat(ENG-13677): Terraform credential helper

tigh-latte:
# Description

Adds a **Terraform credentials helper** for Cloudsmith registries so
`terraform init` can authenticate against a Cloudsmith Terraform registry with
**no token on disk**, reusing the existing Cloudsmith CLI credential chain
(`--api-key`/`CLOUDSMITH_API_KEY`, `credentials.ini`, the OS keyring, or OIDC).

Terraform discovers credentials helpers as executables named
`terraform-credentials-<name>` and only searches its plugin directories (never
`$PATH`). This PR ships the helper, a launcher/wrapper that satisfies that
discovery contract, an installer that wires up `~/.terraformrc`, and the
repository-scoped token format the registry expects.

## What's included

**Command (`cloudsmith credential-helper terraform`)**
- Implements Terraform's credentials-helper protocol for the `get` verb.
  Accepts Terraform's `[verb] <hostname>` calling convention (verb defaults to
  `get`; hostname falls back to stdin) so the launcher can forward args
  verbatim.
- Returns `{"token": "..."}` for a Cloudsmith host and `{}` for any other host
  (exit 0) so Terraform falls back to its own credential sources.
  `store`/`forget` and unknown verbs return an actionable error and a non-zero
  exit. Missing credentials for a Cloudsmith host produce a clean refusal, never
  a traceback.
- Registered on the `credential-helper` group.

**Installer (`credential-helper install/uninstall/list terraform`)**
- Writes the launcher into Terraform's plugin dir (`~/.terraform.d/plugins` by
  default; `--bin-dir` overrides) and manages the `credentials_helper
  "cloudsmith"` block in `~/.terraformrc` via a regex block editor (no HCL
  parser); refuses if a *different* helper block already exists.
- Bakes the resolved `--org`, `-P/--profile`, and `--repo` into the terraformrc
  `args` list so `terraform init` needs no environment variables. Computes the
  terraformrc change before writing the launcher so a conflict leaves no orphan.

**Repository-scoped, custom-domain-aware token format**
- Terraform never tells a credentials helper which repository is requested, so
  `-r/--repo/--repository` (or `CLOUDSMITH_REPO`) is **required** on `get`, and
  can also be passed to `install` to bake it into the terraformrc `args`.
- Tokens are scoped per repository:
  - Standard `*.cloudsmith.io`/`*.cloudsmith.com` hosts → `{org}/{repo}/{token}`
    (the organisation is required here).
  - Custom domains (already bound to a single organisation) → `{repo}/{token}`
    (the org is used only to *resolve* the domain, never emitted in the token).
- Added `is_standard_cloudsmith_domain()` to `credential_helpers/common.py` to
  distinguish standard hosts from custom domains (`is_cloudsmith_domain` was
  refactored to reuse it; behaviour unchanged).

## Type of Change

- [x] New feature

## Testing

- Unit tests for the runtime (`get`/`store`/`forget`, token / empty-object /
  refusal outcomes, org-required-on-standard-domain, org-omitted-on-custom-
  domain), the CLI shim, the wrapper delegation, the terraformrc block
  helpers, and the installer (launcher + terraformrc, `--org`/`-P`/`--repo`
  baking, conflict handling).
- `is_standard_cloudsmith_domain` unit coverage (apex/subdomain/scheme/casing
  matches, plus custom-domain and lookalike non-matches).
- A live integration test (`@pytest.mark.integration`) that runs a real
  `terraform init` in an isolated `HOME` and asserts it does not fail for
  authentication reasons; skips cleanly when `terraform`, the wrapper, or the
  `PYTEST_CLOUDSMITH_*` vars are absent.

```
env -u CLOUDSMITH_PROFILE -u CLOUDSMITH_ORG -u CLOUDSMITH_WORKSPACE \
    -u CLOUDSMITH_KEYRING_BACKEND HOME=$(mktemp -d) \
    pytest -m "not integration"
```

## Additional Notes

- Auth with a JWT is dependent on https://github.com/cloudsmith-io/cloudsmith/pull/16538 being merged and released
  terraformrc block management, the required repo flag, and the token format).


cloudsmith_cli/credential_helpers/terraform/installer.py
```diff
@@ -0,0 +1,289 @@
...
+    return Path.home() / ".terraformrc"
+
+
+def _default_plugin_dir() -> Path:
+    """Return Terraform's conventional user plugin directory.
+
+    ``~/.terraform.d/plugins`` on all platforms — the best-known of Terraform's
+    default credentials-helper search locations and writable without elevation.
+    """
+    return Path.home() / ".terraform.d" / "plugins"
```
cloudsmith-iduffy, line 69: How does this respond on windows? I think window's path is %appdata%/terraform.d/plugins 

https://stackoverflow.com/questions/57742564/terraform-not-finding-third-party-plugin-in-directory-specified-for-plugins#:~:text=Terraform%20looks%20for%20plugins%20in%20a%20number%20of%20locations%2C%20but%20the%20primary%20place%20for%20manually%2Dinstalled%20plugins%20is%20in%20the%20%22User%20Plugins%20Directory%22%2C%20which%20is%20either%20~/.terraform.d/plugins%20on%20Unix%20systems%20or%20%25APPDATA%25%5Cterraform.d%5Cplugins%20on%20Windows.

cloudsmith_cli/credential_helpers/terraform/installer.py
```diff
@@ -0,0 +1,289 @@
...
+        resolved to an absolute path.
+        """
+        if bin_dir is not None:
+            return Path(bin_dir).resolve()
+        return _default_plugin_dir()
+
+    def _launcher_path(self, target_dir: Path) -> Path:
+        """Return the launcher path within *target_dir* for this platform."""
+        if os.name == "nt":
+            return target_dir / f"{self.LAUNCHER_NAME}.cmd"
```
cloudsmith-iduffy: Will .cmd successfully launch on windows as a plugin? https://github.com/hashicorp/terraform/blob/main/internal/providercache/cached_provider.go#L106 looks like the discovery process will get it. Might be worth spinning up a windows instance on aws and testing.

cloudsmith_cli/credential_helpers/terraform/installer.py
```diff
@@ -0,0 +1,289 @@
...
+        ``PATH``. A frozen standalone binary (PyInstaller) is not guaranteed to
+        be on ``PATH`` under that name, so point the launcher at the absolute
+        executable instead. The path is quoted so a directory containing spaces
+        still execs correctly.
+        """
+        if getattr(sys, "frozen", False):
+            return f'"{sys.executable}" credential-helper terraform'
+        return cls.TARGET_CMD
+
+    def _resolve_bin_dir(self, bin_dir: str | None) -> Path:
```
BartoszBlizniak: Terraform searches only its [default plugin locations](https://developer.hashicorp.com/terraform/internals/credentials-helpers#installing-a-credentials-helper). Installation can therefore succeed while Terraform cannot find the helper. Could this be restricted to recognized Terraform plugin directories?

cloudsmith_cli/cli/commands/credential_helper/manage.py
```diff
@@ -65,6 +70,59 @@ def _get_installer(name: str):
     return cls()
 
 
+def _terraform_helper_args(ctx, opts, repo: str | None = None) -> tuple[str, ...]:
```
BartoszBlizniak, line 77: How does this work with `--config-file` / `--credentials-file` flags? If an installation uses either option, `terraform init` may not resolve the same credentials (?)

cloudsmith_cli/cli/commands/credential_helper/terraform.py
```diff
@@ -0,0 +1,137 @@
...
+    elif len(params) == 1:
+        hostname = params[0]
+
+    if not hostname:
+        try:
+            hostname = sys.stdin.read().strip()
+        except (OSError, ValueError):
+            hostname = ""
+
+    exit_code, stdout, stderr = execute(
```
BartoszBlizniak, line 124: The helper reaches [`execute()`](https://github.com/cloudsmith-io/cloudsmith-cli/blob/368bc9a8186b7f2caeff9119456722a666e10c0b/cloudsmith_cli/cli/commands/credential_helper/terraform.py#L124-L132) without reading stdin for `store`. Terraform requires unsupported store operations to [consume the complete payload](https://developer.hashicorp.com/terraform/internals/credentials-helpers#store-store-new-credentials-for-the-given-hostname) before returning an error. Could stdin be drained first?

BartoszBlizniak (changes requested):

cloudsmith-iduffy (approved):

tigh-latte:
@BartoszBlizniak this is ready for another look now

cloudsmith_cli/credential_helpers/terraform/installer.py
```diff
@@ -0,0 +1,492 @@
...
+      (``$XDG_DATA_HOME`` or ``~/.local/share``, then each ``$XDG_DATA_DIRS``
+      entry or the ``/usr/local/share`` and ``/usr/share`` defaults).
+
+    On every OS a ``terraform.d/plugins`` directory in the current working
+    directory is also searched.
+
+    See:
+    https://developer.hashicorp.com/terraform/cli/config/config-file#implied-local-mirror-directories
+    """
+    dirs: list[Path] = [_default_plugin_dir()]
```
BartoszBlizniak: This allowlist is built from the "Implied Local Mirror Directories" doc, but that section describes where Terraform looks for **provider** mirrors, not credentials helpers. Helper discovery is much narrower in the Terraform source:

- [`commands.go#L507-L509`](https://github.com/hashicorp/terraform/blob/main/commands.go#L507-L509): `FindPlugins("credentials", cliconfig.GlobalPluginDirs())`
- [`cliconfig/plugins.go`](https://github.com/hashicorp/terraform/blob/main/internal/command/cliconfig/plugins.go): `GlobalPluginDirs()` returns only `<ConfigDir>/plugins` and `<ConfigDir>/plugins/<GOOS>_<GOARCH>`
- `ConfigDir()` is `~/.terraform.d` on Unix and `%APPDATA%\terraform.d` on Windows ([`config_unix.go`](https://github.com/hashicorp/terraform/blob/main/internal/command/cliconfig/config_unix.go), [`config_windows.go`](https://github.com/hashicorp/terraform/blob/main/internal/command/cliconfig/config_windows.go)). No XDG handling.
- [`discovery/find.go#L54`](https://github.com/hashicorp/terraform/blob/main/internal/plugin/discovery/find.go#L54) does a flat `ReadDir` on each of those dirs. Not recursive.

So as written this still allows installs Terraform will never see:

1. The XDG data dirs, `/usr/local/share` / `/usr/share`, the macOS `io.terraform` dirs and the cwd `terraform.d/plugins` are never searched for helpers. Installing there "succeeds" and `terraform init` silently ignores the helper, which is the failure mode this change was meant to prevent.
2. `_is_recognized_plugin_dir` accepts any descendant of a root, but only `<root>` and `<root>/<os>_<arch>` are searched. `~/.terraform.d/plugins/custom` is not found by Terraform.
3. Line 128 has `"HashCorp"` (should be `HashiCorp`). Moot once the branch goes, but worth noting the list was not verified against a real Terraform.

Suggest reducing the allowlist to exactly `_default_plugin_dir()` and `_default_plugin_dir() / f"{os}_{arch}"`, and dropping the XDG / macOS / cwd branches and their tests. Could you also verify with a real `terraform init` and `TF_LOG=debug` that a launcher in a rejected dir is in fact not picked up, and one in the accepted dirs is?

cloudsmith_cli/cli/tests/test_credential_helper_terraform_installer.py
```diff
@@ -0,0 +1,759 @@
...
+def test_installer_respects_bin_dir_override(tmp_path, monkeypatch):
+    """--bin-dir overrides the default plugin directory for the launcher.
+
+    Uses a subdirectory of the default plugin root — recognized by Terraform on
+    every OS — so the override is honoured without tripping the plugin-directory
+    validation, while proving the launcher lands somewhere other than the root.
+    """
+    monkeypatch.setattr(Path, "home", staticmethod(lambda: tmp_path))
+    monkeypatch.delenv("TF_CLI_CONFIG_FILE", raising=False)
+    custom = tmp_path / ".terraform.d" / "plugins" / "custom"
```
BartoszBlizniak: This test asserts that installing into `~/.terraform.d/plugins/custom` succeeds, but Terraform does a flat `ReadDir` on `plugins/` and `plugins/<os>_<arch>` only, so a launcher here is never discovered. The test currently locks in the bug. Once the allowlist is tightened, the only valid subdirectory to use here is an `<os>_<arch>` one, e.g. `linux_amd64`.

BartoszBlizniak (changes requested):
Two of the three earlier threads (`--config-file`/`--credentials-file` baking and draining stdin on `store`) are fixed correctly, thanks. The plugin-dir restriction needs another pass, see inline.

BartoszBlizniak (approved):

PR #404, opened 2026-09-02
feat(ENG-14143): adding colour suppression to the CLI via environment variables

coillteoir:
---
[//]: # (BEGIN SAPLING FOOTER)
* __->__ #404
* #402


cloudsmith_cli/cli/decorators.py
```diff
@@ -151,6 +153,11 @@ def wrapper(ctx, *args, **kwargs):
 def common_cli_output_options(f):
     """Add common CLI output options to commands."""
 
+    @click.option(
+        "--color",
+        default="auto",
+        type=click.Choice(ColorMode, case_sensitive=False),
```
cloudsmith-iduffy, line 160: Can we move the envvar into the click option instead of doing it as os.environ?

Adding `envvar="color"` should do it https://click.palletsprojects.com/en/stable/options/#values-from-environment-variables

cloudsmith-iduffy (changes requested):

cloudsmith_cli/cli/decorators.py
cloudsmith-iduffy: @copilot can you take this on and bring it to resolution? :) we ideally want no raw os.environ calls, it should all be done via click options.

cloudsmith-iduffy (approved):

PR #416, opened 2026-09-08
release: cloudsmith-cli v1.27.0

cloudsmith-iduffy:
# Description

Prepare Cloudsmith CLI 1.27.0 from the changes already merged into master since v1.26.0.

- Bump the release version to 1.27.0 in `.bumpversion.cfg` and `cloudsmith_cli/data/VERSION`.
- Finalize the 1.27.0 changelog to describe the behavior already present in master.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [x] Other (release)


packaging/smoketest.sh
```diff
@@ -178,6 +178,47 @@ run_offline() {
...
+  no_dep_error "$OUT" "credential-helper install terraform"
+  case "$BIN" in
+    *.exe)
+      TERRAFORM_HELPER="$TERRAFORM_TMP/appdata/terraform.d/plugins/terraform-credentials-cloudsmith.exe"
+      ;;
+    *)
+      TERRAFORM_HELPER="$TERRAFORM_TMP/home/.terraform.d/plugins/terraform-credentials-cloudsmith"
+      ;;
+  esac
+  [ -x "$TERRAFORM_HELPER" ] \
```
cloudsmith-iduffy: The Windows offline smoketest looks for the launcher in a different directory than the one the install actually wrote to.

`TERRAFORM_TMP` comes from `mktemp -d` under Git Bash, so it is an MSYS path such as `/tmp/tmp.abc`. MSYS converts POSIX paths in *arguments* to native paths, but not in exported environment variables, so `cloudsmith.exe` receives `APPDATA=/tmp/tmp.abc/appdata` verbatim. `_default_plugin_dir()` then resolves it with `pathlib` as `<current drive>\tmp\tmp.abc\appdata\terraform.d\plugins`, while bash resolves `$TERRAFORM_TMP/appdata/...` through the MSYS `/tmp` mount (`<git-root>\tmp\...`). The two never agree, so this line fails with `terraform install did not create an executable helper` even when the install succeeded.

Convert the paths for the child process, e.g. `TERRAFORM_TMP_WIN=$(cygpath -w "$TERRAFORM_TMP")` and pass that as `APPDATA`/`HOME`/`TF_CLI_CONFIG_FILE`, or read the installed path out of `$OUT` (the installer already prints `wrote launcher <path>`) instead of recomputing it in the shell.

cloudsmith_cli/credential_helpers/terraform/installer.py
cloudsmith-iduffy, line 1: `_install_exe_launcher` copies the frozen `.exe` out of its PyInstaller bundle, which makes it unrunnable.

`cloudsmith.spec` builds a **onedir** bundle (`EXE(..., exclude_binaries=True)` + `COLLECT`). A onedir executable locates its payload (`_internal`/`base_library.zip`, the Python DLL, the collected binaries) relative to its own directory. `_resolve_helper_exe()` prefers the bundled `terraform-credentials-cloudsmith.exe` whenever `sys.frozen` is set, and `_install_exe_launcher` then `shutil.copy2()`s that single file into `%APPDATA%\terraform.d\plugins`. The copy has no payload beside it, so it fails to start ("Failed to load Python DLL" / "could not find base_library.zip"), and `terraform init` sees a helper that cannot run.

Concrete scenario: a Windows user unzips the standalone build and runs `cloudsmith credential-helper install terraform`. The install reports success, `terraform init` then fails on every Cloudsmith registry. The new `packaging/smoketest.sh` step in this PR executes the *installed* helper, so the `windows-x86_64` build job should reproduce this.

Options: build the terraform helper as a self-contained onefile `EXE` (its own `EXE(..., exclude_binaries=False)` target) so it stays runnable when relocated, or skip the frozen bundle in `_resolve_helper_exe` and fall back to a `pip`-generated `Scripts\terraform-credentials-cloudsmith.exe` (that launcher embeds an absolute interpreter path and does survive being copied).

cloudsmith_cli/credential_helpers/terraform/installer.py
```diff
@@ -195,15 +195,28 @@ class TerraformInstaller:
...
+            return f"{shlex.quote(sys.executable)} credential-helper terraform"
+
+        scripts_dir = Path(sys.executable).parent
+        helper = scripts_dir / cls.LAUNCHER_NAME
+        if helper.is_file():
+            return shlex.quote(str(helper))
+
+        cloudsmith = scripts_dir / "cloudsmith"
+        if cloudsmith.is_file():
+            return f"{shlex.quote(str(cloudsmith))} credential-helper terraform"
```
cloudsmith-iduffy: This drops the `PATH` fallback, so `install terraform` now hard-fails on any install whose console scripts do not sit beside `sys.executable`.

`pip install --user`, a Homebrew/system Python with `--break-system-packages`, and Debian`s `python3-*` layout all put `cloudsmith` in `~/.local/bin`, `~/Library/Python/3.x/bin` or `/usr/local/bin` while `sys.executable` stays `/usr/bin/python3`. Neither `scripts_dir / LAUNCHER_NAME` nor `scripts_dir / "cloudsmith"` exists, so `TerraformHelperExeNotFound` is raised and the install refuses even though `cloudsmith` runs fine.

Keeping the sibling scripts as the preferred answer and only searching `PATH` as a last resort preserves the intent (the launcher still records an absolute file, not a bare command name) without the hard failure:

```suggestion
        scripts_dir = Path(sys.executable).parent
        for candidate in (
            scripts_dir / cls.LAUNCHER_NAME,
            shutil.which(cls.LAUNCHER_NAME),
        ):
            if candidate and Path(candidate).is_file():
                return shlex.quote(str(candidate))

        for candidate in (scripts_dir / "cloudsmith", shutil.which("cloudsmith")):
            if candidate and Path(candidate).is_file():
                return f"{shlex.quote(str(candidate))} credential-helper terraform"
```

cloudsmith_cli/credential_helpers/terraform/terraformrc.py
cloudsmith-iduffy, line 1: `remove_block` and `TerraformInstaller.status` both use `find_block`, which returns only the *first* `credentials_helper` match.

Concrete scenario: the user installs the Cloudsmith helper, then hand-adds a `credentials_helper "vault"` block *above* it. `cloudsmith credential-helper uninstall terraform` reports `no credentials_helper "cloudsmith" block to remove` and leaves the Cloudsmith block in place, even though it removes the launcher — so the terraformrc keeps pointing Terraform at an executable that no longer exists. `status` reports the helper as not configured for the same reason.

`_find_conflicting_block` already does the right thing with `finditer`; the removal/status path should scan the same way:

```suggestion
    # Scan every match rather than only the first: a foreign block earlier in
    # the file must not hide the Cloudsmith block from removal.
    match = next(
        (m for m in _BLOCK_RE.finditer(text) if m.group("name") == HELPER_NAME), None
    )
    if match is None:
        return text, False
```

cloudsmith_cli/cli/tests/test_credential_helper_terraform.py
```diff
@@ -499,6 +512,18 @@ def test_cli_prints_empty_object_for_foreign_host(runner):
 
     assert result.exit_code == 0
```
cloudsmith-iduffy, line 500: The `assert "k_abc" not in result.stdout` line landed in the new test instead of staying with this one, so `test_cli_prints_empty_object_for_foreign_host` no longer checks what its docstring promises ("no leaked token"). Add it back here as well — it is the assertion that catches a regression printing the raw API key for a foreign host.

```suggestion
    assert result.exit_code == 0
    assert json.loads(result.stdout) == {}
    assert "k_abc" not in result.stdout
```

apoclyps (approved):

PR #419, opened 2026-09-15
feat(ENG-14145): self-update command and background update check

tigh-latte:
## Summary

Adds a `cloudsmith update` command for self-updating standalone binary installs, plus a non-intrusive background check that notifies users when a newer version is available.

## Changes

- **`update` command** (`cli/commands/update.py`): self-updates the standalone binary installation. Reads the release manifest, downloads and replaces target files in place (per-file replace rather than swapping the whole install dir, so user files in the install dir are preserved). Handles Windows update quirks.
- **Background update check** (`core/update_check.py`): probes for the latest version of the *host's own* platform/arch (previously always probed `linux-x86_64-gnu`). Uses a non-configurable manifest URL. Tracks a second "last notified" timestamp so a notification can be suppressed until a later run (e.g. when running in JSON mode).
- **Install detection** (`core/installation.py`): detects how the CLI was installed to decide whether self-update applies.
- **Self-update engine** (`core/self_update.py`): download/verify/replace logic.
- Config plumbing in `cli/config.py` + `data/config.ini`, and wiring into `main.py` / `registry.py` / `utils.py`.

## Tests

New unit tests for the update command, self-update engine, installation detection, background check, and main-group integration.

## Notes

Tracking: ENG-14145

### DIFF
--- a/cloudsmith_cli/cli/commands/update.py
+++ b/cloudsmith_cli/cli/commands/update.py
@@ -1,0 +1,157 @@
+"""CLI/Commands - Update the CLI to the latest released version."""
+
+import click
+
+from ...core import installation, self_update, update_check
+from ...core.session import create_requests_session
+from ...core.version import get_version, parse_version
+from .. import decorators, utils
+from ..utils import maybe_spinner
+from .main import main
+
+
+def _fetch_manifest(opts, session, target):
+    import requests
+
+    try:
+        with maybe_spinner(opts):
+            return update_check.fetch_latest_manifest(session, target=target)
+    except (requests.RequestException, ValueError) as exc:
+        raise click.ClickException(f"Failed to fetch the latest version: {exc}")
+
+
+def _is_up_to_date(latest, current):
+    try:
+        return parse_version(latest) <= parse_version(current)
+    except ValueError as exc:
+        raise click.ClickException(f"Cannot compare versions: {exc}")
+
+
+def _confirm_self_update(opts, current, latest, assume_yes):
+    """Return True to proceed with the self-update.
+
+    Auto-confirms when ``--yes`` is given, when output is machine-readable, or
+    when stderr is not a terminal (so scripted/piped invocations do not hang on
+    a prompt).
+    """
+    if assume_yes or utils.should_use_stderr(opts) or not update_check.stderr_is_tty():
+        return True
+    return click.confirm(
+        f"Update the Cloudsmith CLI from {current} to {latest}?", default=True
+    )
+
+
+def _run_self_update(opts, session, manifest, data):
+    import requests
+
+    use_stderr = utils.should_use_stderr(opts)
+    click.echo(
+        f"Downloading and installing version {manifest['version']} ... ",
+        nl=False,
+        err=use_stderr,
+    )
+    try:
+        with maybe_spinner(opts):
+            self_update.perform_self_update(manifest, session=session)
+    except (self_update.SelfUpdateError, requests.RequestException, OSError) as exc:
+        click.secho("ERROR", fg="red", err=use_stderr)
+        raise click.ClickException(str(exc))
+    click.secho("OK", fg="green", err=use_stderr)
+    data["upgraded"] = True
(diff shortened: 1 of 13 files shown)


cloudsmith_cli/cli/commands/update.py
```diff
@@ -0,0 +1,157 @@
...
+    treat "update available, self-update unsupported" as a failure.
+    """
+    data["outcome"] = "manual"
+    data["download_url"] = manifest.get("url")
+    data["sha256"] = manifest.get("sha256")
+    data["archive"] = manifest.get("archive")
+    if utils.maybe_print_as_json(opts, data):
+        return
+    click.echo(
+        f"A new version of the Cloudsmith CLI is available: {current} \u2192 {latest}"
```
cloudsmith-iduffy, line 82: It would be good to include a link here to the change log to easily allow the user to know what they will gain by upgrading.

tigh-latte: Good idea

cloudsmith_cli/core/self_update.py
```diff
@@ -0,0 +1,382 @@
...
+                moved.append((target, None))
+            _log_rename(incoming, target)
+        # Defensive re-check: the staged bundle was validated above.
+        if executable_name is not None and not os.path.isfile(
+            os.path.join(install_dir, executable_name)
+        ):
+            raise SelfUpdateError(
+                f"the installed bundle is missing {executable_name} after replace"
+            )
+    except (OSError, SelfUpdateError) as exc:
```
cloudsmith-iduffy: Will we land in here if a user does ctrl+c mid upgrade? We should likely do something here to ensure we still do the rollback / clean up on a interrupt.

cloudsmith_cli/cli/commands/update.py
```diff
@@ -0,0 +1,164 @@
...
+    # pylint: disable=unused-argument
+    current = get_version()
+    channel = installation.detect_channel()
+    target = installation.detect_target()
+    if channel == installation.CHANNEL_STANDALONE and target is None:
+        raise click.ClickException(
+            "Cannot detect a supported platform for the standalone binary."
+        )
+
+    session = create_requests_session(user_agent=opts.api_user_agent)
```
cloudsmith-iduffy: If a user has HTTP_PROXY / HTTPS_PROXY set/configured will the upgrade requests session use it accordingly? I believe it will.

However, if a user has api_proxy set, then we should likely not funnel those requests over that proxy given its "api_" prefix. 

Not for this PR but in the V2 release we should likely look towards getting rid of `api_proxy` and just using HTTP_PROXY / HTTPS_PROXY everywhere.

cloudsmith_cli/core/update_check.py
```diff
@@ -0,0 +1,477 @@
...
+    """
+    invoked = ctx.invoked_subcommand
+    invoked = getattr(ctx.command, "inverse", {}).get(invoked, invoked)
+    if update_check_disabled(
+        no_check_flag=no_check_flag, config_value=opts.check_for_update
+    ):
+        return
+
+    thread: threading.Thread | None = None
+    if is_check_due(read_last_check_time()):
```
cloudsmith-iduffy: What happens here if I'm running 1.26.0, 1.27.0 is released and I get that update notification and its cached, but then 1.28.0 is released I don't see how we drop into the else branch here to do the refresh?

cloudsmith_cli/core/self_update.py
tigh-latte: Good spot, I've removed the narrow error handling and instead will do a rollback on any error from here.

cloudsmith_cli/core/update_check.py
tigh-latte: addresseed in [4f11e47](https://github.com/cloudsmith-io/cloudsmith-cli/pull/419/changes/4f11e4732cd2ae1701b52103fdef043b9aaf575a)

cloudsmith-iduffy (approved):
Ship it in sync with docs! :D

